PR #3, opened 2021-11-03
Feature/gradient descent

campsd:


funfact/lang/interpreter/_index_propagation.py
```diff
@@ -51,16 +51,16 @@ def neg(self, x: Numeric, **kwargs):
 
     @as_payload
     def mul(self, lhs: Numeric, rhs: Numeric, **kwargs):
-        return set(lhs.live_indices).symmetric_difference(rhs.live_indices)
+        return list(set(lhs.live_indices).symmetric_difference(rhs.live_indices))
```
yhtang, line 54: Could you please dig out the old implementation, and wrap it as a method called `ordered_symmetric_difference`, so that we can preserve the 'natural' order of the indices?

funfact/lang/interpreter/_initialization.py
```diff
@@ -1,12 +1,29 @@
...
 from ._base import TranscribeInterpreter
 
 
+
+class JaxRng:
+    
+    def __init__(self, key=0):
+        self.key = jax.random.PRNGKey(key)
+        
+    def rand(self, size, dtype=np.float32):
```
yhtang, line 16: I'd suggest we call this method `normal` to be consistent with the underlying JAX method.

funfact/optim/_adam_np.py
```diff
@@ -0,0 +1,27 @@
...
+        self.lr = lr
+        self.beta1 = beta1
+        self.beta2 = beta2
+        self.epsilon = epsilon
+        self.M = [np.zeros(w.shape) for w in x.factors]
+        self.V = [np.zeros(w.shape) for w in x.factors]
+
+    def step(self, grad):
+        grad_factors = grad.factors
+        new_factors = self.x.factors
```
yhtang, line 20: I'd suggest we keep the details of `Factorization` out of here, i.e., so that `x` is already a list of tensors, and so is `grad`. This should give better decoupling between the modules.

yhtang (changes requested):

funfact/optim/_adam_np.py
yhtang: After a second thought, I realize that we need to assign to the `.factors` attribute. Let me check how the JAX NN modules handle this situation.

yhtang:
Cool! Could you please fix the Lint errors? We should be able to merge after that's nailed. Thanks!

yhtang (approved):

PR #16, opened 2021-11-07
WIP: explicit specification of lhs in tensor expressions

### LINKED ISSUE #6: Explicit specification of tensor expression output indices
The `einsum` API of common numerical linear algebra libraries such as NumPy, PyTorch, and JAX all support the extended Einstein notation format, which allows an explicit expression of the left-hand side indices in the specs string. This explicit specification is the part after the `->` symbol:
```
# implicit lhs, inferred to be ik
ij,jk

# explicit lhs
ij,jk->ik

# explicit lhs, result transposed
ij,jk->ki

# explicit lhs, disables summation over j
ij,jk->ijk
```
As such, the specification of left-hand side indices has the effect of transposition and explicitly enabling/disabling the reduction operation over certain indices.

Adding this functionality would entail semantics that may look like:
```
# A, B, C, D: tensors
# i, j, k: indices
C[k, i] = A[i, j] * B[j, k]
D[i, k, j] = A[i, j] * B[j, k]
```

yhtang:
Added the `==` operator for the explicit specification of output tensor indices.

For example,
```
A = ff.tensor(2, 3)
B = ff.tensor(3, 4)
C = ff.tensor(2, 4)
tsrex = C[k, i] == A[i, j] * B[j, k]
```
give rise to the following tensor expression
```
 ==
 ├── k
 ├── i
 ╰── *
     ├── A[i,j]
     │   ├── A
     │   ├── i
     │   ╰── j
     ╰── B[j,k]
         ├── B
         ├── j
         ╰── k
```
which in addition to carrying out the inner product, also transposes the final result.

Note that this PR alone does not fully implement the `einsum` semantics, where an index in the output string also suppresses the reduction over its associated dimensions. This will be address by #17 through the introduction of an **index modifier** syntax.

Closes #6 upon completion.

### DIFF
--- a/funfact/lang/_ast.py
+++ b/funfact/lang/_ast.py
@@ -50,35 +50,42 @@ class Primitives:
         tensor: _ASNode,
         indices: Union[AbstractIndex, Iterable[AbstractIndex]]
     ):
-        '''indexed notation for a single tensor'''
+        '''indexed notation for a single tensor: tensor[indices...]'''
 
     @primitive(precedence=2)
     def call(f: str, x: _ASNode):
-        '''nonlinear function call'''
+        '''nonlinear function call: f(x)'''
 
     @primitive(precedence=3)
     def pow(base: _ASNode, exponent: _ASNode):
-        '''raise to power'''
+        '''raise to power: base**exponent'''
 
     @primitive(precedence=4)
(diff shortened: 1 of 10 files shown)


yhtang:
Superseded by #19.

PR #18, opened 2021-11-08
Feature/tsrex func

### LINKED ISSUE #5: Creating abstract tensors from concrete ones

campsd:
Resolves #5


funfact/lang/_tsrex.py
```diff
@@ -130,16 +130,24 @@ def tensor(*spec, initializer=None):
...
     '''
     if len(spec) == 2 and isinstance(spec[0], str) and _is_tensor(spec[1]):
-        raise NotImplementedError()
-    elif len(spec) == 1 and _is_tensor(spec[1]):
-        raise NotImplementedError()
+        symbol = spec[0]
+        initializer = spec[1]
+        size = initializer.shape
+    elif len(spec) == 1 and _is_tensor(spec[0]):
+        symbol = f'Anonymous_{AbstractTensor.n_nameless}'
```
yhtang, line 137: Need to increment `AbstractTensor.n_nameless` after using it just like a few lines down:
```
AbstractTensor.n_nameless += 1
```

funfact/lang/_tsrex.py
```diff
@@ -130,16 +130,24 @@ def tensor(*spec, initializer=None):
...
+        size = initializer.shape
     elif isinstance(spec[0], str):
         symbol, *size = spec
     else:
         symbol = f'Anonymous_{AbstractTensor.n_nameless}'
         AbstractTensor.n_nameless += 1
         size = spec
 
+    if hasattr(size[0], "__len__"):
+        initializer = size[0]
```
yhtang, line 148: I'm a bit confused as to why `size[0]` could serve as the initializer.

funfact/lang/interpreter/_initialization.py
```diff
@@ -30,6 +30,8 @@ def scalar(self, value, **kwargs):
     @as_payload
     def tensor(self, value, **kwargs):
         if value.initializer is not None:
+            if not callable(value.initializer):
+                return value.initializer
```
yhtang, line 34: We'll probably need to return a copy of the initializer if it is a tensor so that the original value would not be changed by any following operations.

funfact/util/typing.py
```diff
@@ -2,5 +2,5 @@
 # -*- coding: utf-8 -*-
 
 
-def _is_tensor():
-    raise NotImplementedError()
+def _is_tensor(x):
```
yhtang, line 5: Things like nested lists could also qualify under this criterion. Could we learn from how other NLA libraries are doing this? I think `np.array` or `np.asarray` may perform this kind of check.

yhtang (changes requested):

funfact/lang/_tsrex.py
campsd: I have removed this, was from a previous version and should have been removed after rebasing

yhtang (approved):

Issue #17, opened 2021-11-07
Explicit suppression of reduction over select indices


yhtang:
Candidate syntax:
```
(A[i, k] * B[k, j]) @ (i, j, k)

(A[i, ~k] * B[~k, j])
```

yhtang:
Closed by #21.

PR #19, opened 2021-11-08
Added the `>>` syntax for specifying output indices.

### LINKED ISSUE #6: Explicit specification of tensor expression output indices

yhtang:
Summary:
- Added the `>>` operator for the explicit specification of output tensor indices.
- A shorthand syntax, which introduces the `~` modifier to individual indices, will be implemented in #17.
- Replaces #16.
- Closes #6.

Example:
```
A = ff.tensor('A', 2, 3)
B = ff.tensor('B', 3, 4)
tsrex = A[i, j] * B[j, k] >> (k, j)
```
give rise to the following tensor expression
```
 >>
 ├── *
 │   ├── A[i,j]
 │   │   ├── A
 │   │   ╰── i,j
 │   │       ├── i
 │   │       ╰── j
 │   ╰── B[j,k]
 │       ├── B
 │       ╰── j,k
 │           ├── j
 │           ╰── k
 ╰── k,j
     ├── k
     ╰── j
```
which in addition to carrying out the inner product, also transposes the final result.

This is currently a **work-in-progress**: a forward-backward interpreter is needed to implement index analysis for the new syntax.

### DIFF
--- a/funfact/lang/_semiring.py
+++ b/funfact/lang/_semiring.py
@@ -1,0 +1,10 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+from ._ast import Primitives as P
+from ._tsrex import TsrEx, EinopEx
+
+
+def minplus(lhs: TsrEx, rhs: TsrEx):
+    return EinopEx(P.ein(
+        lhs._as_node(lhs), rhs._as_node(rhs), 6, 'min', 'add', None
+    ))
--- a/funfact/lang/_tensor.py
+++ b/funfact/lang/_tensor.py
@@ -17,16 +17,20 @@ class Identifier(ABC):
     @symbol.setter
     def symbol(self, string: str):
         m = re.fullmatch(r'([a-zA-Z]+)(?:_(\d+))?', string)
-        if m is None:
-            raise RuntimeError(
-                f'{repr(string)} is not a valid symbol.\n'
-                'A symbol must be alphabetic and optionally followed by an '
-                'underscore and a numeric subscript. '
-                'Examples: i, j, k_0, lhs, etc.'
-            )
-        self._letter, self._number = m.groups()
-        if self._letter == 'Anonymous':
-            self._letter = r'\varphi'
+        try:
+            self._letter, self._number = m.groups()
+        except AttributeError:
+            m = re.fullmatch(r'__(\d+)', string)
+            try:
+                self._letter = r'\lambda'
+                self._number, = m.groups()
+            except AttributeError:
+                raise RuntimeError(
+                    f'{repr(string)} is not a valid symbol.\n'
+                    'A symbol must be alphabetic and optionally followed by '
+                    'an underscore and a numeric subscript. '
+                    'Examples: i, j, k_0, lhs, etc.'
+                )
 
     @abstractmethod
     def _repr_tex_(self):
--- a/funfact/lang/interpreter/_einop.py
+++ b/funfact/lang/interpreter/_einop.py
@@ -5,7 +5,17 @@ import numpy
 import re
 
 
-def _einop(spec: str, lhs, rhs, op):
+class DummyBackend:
+
+    add = np.add
+    sub = np.subtract
+    mul = np.multiply
+    div = np.divide
+    min = np.min
+    sum = np.sum
+
+
+def _einop(spec: str, lhs, rhs, reduction: str, pairwise: str):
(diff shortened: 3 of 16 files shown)


funfact/lang/_tensor.py
```diff
@@ -17,16 +17,20 @@ def symbol(self):
...
-                'underscore and a numeric subscript. '
-                'Examples: i, j, k_0, lhs, etc.'
-            )
-        self._letter, self._number = m.groups()
-        if self._letter == 'Anonymous':
-            self._letter = r'\varphi'
+        try:
+            self._letter, self._number = m.groups()
+        except AttributeError:
+            m = re.fullmatch(r'__(\d+)', string)
```
campsd, line 23: Why do you check for symbols with double underscores? Is this related to #4 ?

campsd (commented):
In the `_index_propagation` interpreter the binary nodes should no longer just return the symmetric difference but use the information from the `let` node (if present).

funfact/lang/_tensor.py
yhtang: Right. This is still a temporary solution. The plan is to allow truly anonymous indices, and then only assign readable letters to them when rendering in latex/ascii.

yhtang:
> In the `_index_propagation` interpreter the binary nodes should no longer just return the symmetric difference but use the information from the `let` node (if present).

Yes. I'm working on an update which actually removes the `let` primitive, merged all the binary primitives as a single `ein` primitive, and make the output indices a builtin field of `ein`. This should eliminate the need for a backward interpreter when doing index/einspec analysis, and also allow future generalization of the binary operators to include things such as the minplus semiring.

funfact/lang/interpreter/_einop.py
```diff
@@ -67,7 +79,11 @@ def _einop(spec: str, lhs, rhs, op):
     con_ax = tuple(con_ax)
 
     # compute the contraction in alphabetical order
-    result = np.sum(op(lhs[dim_lhs], rhs[dim_rhs]), axis=con_ax)
+    op_redu = getattr(DummyBackend, reduction)
+    op_pair = getattr(DummyBackend, pairwise)
+
+    print('op_redu', op_redu)
```
campsd, line 85: Should this be removed?

funfact/lang/_semiring.py
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+from ._ast import Primitives as P
+from ._tsrex import TsrEx, EinopEx
+
+
+def minplus(lhs: TsrEx, rhs: TsrEx):
```
campsd, line 7: This errors for me: 

`tsrex = minplus(A[i, j], B[j,k])` returns
`AttributeError: 'TsrEx' object has no attribute '_as_node'`

funfact/lang/_semiring.py
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+from ._ast import Primitives as P
+from ._tsrex import TsrEx, EinopEx
+
```
campsd, line 5: We can add support for log-spaced Einsum contractions.
We need  to add a reduction of the form shown below to the `DummyBackend` in `_einop.py` and create
an `EinopEx` for this.

```
def logspaceSum(data, axis=None):
    return np.log(np.sum(np.exp(data), axis=axis))
```

funfact/lang/_semiring.py
campsd: The same error happens with the math operations in `_math.py` for which the `_as_node` function is used.

PR #33, opened 2021-11-21
Use the `[[*i, ...]]` syntax to denote Kronecker product indices.

yhtang:
Note: this is only the 'front end' of the Kronecker product syntax. Changed are needed in the evaluator/einop back end to carry out the calculations.

If the same index shows up between a pair of contracting tensors **and** at least one of the occurrences is decorated with the unary index multiplication sign `*`, the a Kronecker-style outer op operation will happen along the associated dimensions.

Example (note the double bracket `[[]]`, which is required in this case since we are _abusing_ the iterable unpacking operator here):
```
A = ff.tensor('A', 4, 4)
B = ff.tensor('B', 3, 4)
i, j = ff.indices('i, j')
A[[*i, j]] * B[i, ~j]
```
Leads to:
```
 sum:mul 
 ├── A[*i,j] 
 │   ├── A 
 │   ╰── *i,j 
 │       ├── *i 
 │       ╰── j 
 ╰── B[i,~j] 
     ├── B 
     ╰── i,~j 
         ├── i 
         ╰── ~j 
```
and
image

This should yield a 12-by-4 matrix, where Kronecker product occurs along the first dimension, and elementwise product occurs along the second dimension.

### DIFF
--- a/funfact/lang/interpreter/_ascii.py
+++ b/funfact/lang/interpreter/_ascii.py
@@ -17,9 +17,11 @@ class ASCIIRenderer(TranscribeInterpreter):
         return str(abstract.symbol)
 
     @as_payload
-    def index(self, item, bound, **kwargs):
+    def index(self, item, bound, kron, **kwargs):
         if bound:
             return f'~{str(item.symbol)}'
+        elif kron:
+            return f'*{str(item.symbol)}'
         else:
             return str(item.symbol)
 
--- a/funfact/lang/_ast.py
+++ b/funfact/lang/_ast.py
@@ -43,8 +43,10 @@ class Primitives:
         '''an abstract tensor'''
 
     @primitive(precedence=0)
-    def index(item: AbstractIndex, bound: bool):
-        '''an index; bound indices are not reduced even if appear twice'''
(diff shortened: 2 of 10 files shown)


funfact/lang/interpreter/_ascii.py
```diff
@@ -17,9 +17,11 @@ def tensor(self, abstract, **kwargs):
         return str(abstract.symbol)
 
     @as_payload
-    def index(self, item, bound, **kwargs):
+    def index(self, item, bound, kron, **kwargs):
```
campsd, line 20: Can an index be simultaneously `bound` and `kron`?

yhtang: Not really, since they represent conflicting types of operations. We can add a checker for this in the syntax validator.

yhtang:
> Looks good, my only question is if we should support indices that are both `bound` and `kron`. But probably that case doesn't make sense though as they are both elementwise operations but with a different axis ordering.

Yeah, `~` leads to elementwise operations while `*` leads to pairwise operations, so they are not compatible.

Issue #14, opened 2021-11-05
Multi-replica factorization


yhtang:

yhtang:
@campsd pointed out the `vmap` pattern as a viable design approach.

campsd:
We can implement this with a vectorizing interpreter that adds a non-reducing dimension at the end of all tensors and index_notations in the AST.

PR #39, opened 2021-11-22
multi replica

campsd:
Draft pull request working towards #14. 
The main outstanding issue is that the vectorizing index is not necessarily the last index after an Einstein operation.

### DIFF
--- a/funfact/lang/interpreter/_vectorize.py
+++ b/funfact/lang/interpreter/_vectorize.py
@@ -8,60 +8,67 @@ from funfact.lang._terminal import AbstractIndex, AbstractTensor, LiteralValue
 from ._base import _deep_apply, TranscribeInterpreter
 
 
-class Vectorizer():
+class Vectorizer(TranscribeInterpreter):
 
     Tensorial = TranscribeInterpreter.Tensorial
     Numeric = TranscribeInterpreter.Numeric
 
-    def literal(self, value: LiteralValue, **kwargs):
-        pass
+    as_payload = TranscribeInterpreter.as_payload('outidx')
 
+    def __init__(self, replicas: int):
+        self.replicas = replicas
+        self.vec_index = P.index(AbstractIndex(), bound=False, kron=False)
+
+    @as_payload
+    def literal(self, value: LiteralValue, **kwargs):
+        return []
+
+    @as_payload
     def tensor(self, abstract: AbstractTensor, **kwargs):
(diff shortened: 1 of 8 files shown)


yhtang:
Great!

> The main outstanding issue is that the vectorizing index is not necessarily the last index after an Einstein operation.

Can we solve it using the trick of directly setting `outidx` using the reordered live indices?

funfact/lang/interpreter/_vectorize.py
```diff
@@ -0,0 +1,67 @@
...
+# -*- coding: utf-8 -*-
+import copy
+from typing import Optional
+from funfact.util.iterable import as_tuple
+from funfact.lang._ast import _AST, _ASNode, Primitives as P
+from funfact.lang._terminal import AbstractIndex, AbstractTensor, LiteralValue
+from ._base import _deep_apply, TranscribeInterpreter
+
+
+class Vectorizer():
```
yhtang, line 11: This class can subclass TranscribeInterpreter, right? Then we don't need to redefine `__call__` again.

yhtang (changes requested):

campsd:
Close #14.
I you want to vectorize a tensor expression, you have to run the index propagation before **and** after the vectorizer. The leaf initializer should only be run **after** vectorizing.

campsd:
The vectorizer now ensures:

- it doesn't modify the properties (shapes, ...) of the original tensor expression
- the vectorizing index is always last in a tensor expression

Example code snippet:
```
T = ff.tensor('T', 3, 3, 3)
u1 = ff.tensor('u_1', 10, 3)
u2 = ff.tensor('u_2', 20, 3)
u3 = ff.tensor('u_3', 30, 3)
i1, i2, i3, k1, k2, k3 = ff.indices('i_1, i_2, i_3, k_1, k_2, k_3')
tsrex = T[k1, k2, k3] * u1[i1, k1] * u2[i2, k2] * u3[i3, k3]
fac = Factorization(tsrex,nvec=4)
fac().shape
```

will return `(10, 20, 30, 4)` and `tsrex` will not be changed.

yhtang (approved):

yhtang:
Somehow, the history of this PR is a bit weird. For example, it even tries to update `CHANGELOG.md` which appears empty here but is actually populated upstream. I'm not sure what has gone wrong. :rofl:

yhtang (approved):

PR #40, opened 2021-11-23
Shape analyzer and exposing properties of tensor expressions

### LINKED ISSUE #31: Exposing properties of the underlying AST for tensor expressions
Certain properties of the AST can be of good practical value for users, yet these are only accessible through a long detour at the moment. For example, the set of 'live' indices, the shape of the resulting tensor, etc. It would be very helpful to expose them as `@property` of the `TsrEx` class.

Properties to expose:
- set of live indices
- shape
- ndim
- einspec
- ...

campsd:
Closes #31.

Example usage:
```
A = ff.tensor('A', 2, 3, 4)
B = ff.tensor('B', 4, 3, 5)
tsrex = A[i, j, k] * B[k, ~j, l]
f = Factorization(tsrex)
print(f.shape)
print(tsrex.shape)
print(tsrex.ndim)
print(tsrex.live_indices)
print(tsrex.einspec)
```

### DIFF
--- a/funfact/lang/_tsrex.py
+++ b/funfact/lang/_tsrex.py
@@ -82,8 +82,14 @@ class _BaseEx(_AST):
     _index_propagator = IndexPropagator()
     _shape_analyzer = ShapeAnalyzer()
 
+    def __init__(self, data=None):
+        super().__init__(data)
+        self.root = self._einspec_generator(self._shape_analyzer(
+                    self._index_propagator(self.root)))
+        self._latex = self._latex_intr(self.root)
+
     def _repr_html_(self):
-        return f'''$${self._latex_intr(self.root)}$$'''
+        return f'''$${self._latex}$$'''
 
     @property
     def asciitree(self):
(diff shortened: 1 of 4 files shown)


funfact/lang/interpreter/_shape.py
```diff
@@ -0,0 +1,79 @@
...
+            else:
+                shape.append(dict_rhs[i])
+        return tuple(shape)
+
+    @as_payload
+    def tran(self, src: Numeric, live_indices, **kwargs):
+        dict_src = dict(zip(src.live_indices, src.shape))
+        shape = []
+        for i in live_indices:
+            shape.append(dict_src[i])
```
yhtang, line 78: Could we collapse line 75-79 to something like this below? :grin: 
```
return [src.shape[src.live_indices.index(i)] for i in live_indices]
```

yhtang (changes requested):

funfact/lang/interpreter/_shape.py
campsd: Done.

funfact/lang/_tsrex.py
```diff
@@ -85,6 +89,24 @@ def _repr_html_(self):
     def asciitree(self):
         return self._asciitree_factory(self.root)
 
+    @property
```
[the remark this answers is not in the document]
campsd: Done. :)

yhtang (approved):

Issue #34, opened 2021-11-21
Backend for Kronecker product operations


yhtang:
This is a follow-up to #33. It seems like both the index propagator, slice propagator, and einop method needs to be updated accordingly. Meanwhile, given that there are now three types of indices (live, keep, and kron) in the einop operation, it might make sense to bypass the einspec procedure and directly use the live indices to carry out einop operations.

campsd:
_Blow up_ Kron indices inside `einop` (using kronecker product with one vector).

PR #49, opened 2021-11-30
Kronecker backend

campsd:
Working on Kronecker backend #34. 

There are a few outstanding issues for this draft PR, but the basic functionality works.

**Issues:**
- `_einop` still uses a spec string for now, everything after the identifier `|` at the end of the spec string is considered a Kronecker index
- there is an issue with the `implied_survival` indices: the order between `keep_bound` and `keep_kron` is arbitrary and can be incorrect. See second test case below.
- the shape analyzer is not yet updated

The code snippet below tests the functionality:
```Python
import funfact as ff
from funfact.lang.interpreter import EinsteinSpecGenerator, IndexPropagator, Evaluator, LeafInitializer
_my_ein = EinsteinSpecGenerator()
_my_index = IndexPropagator()
_my_eval = Evaluator()
_my_init = LeafInitializer()
A = ff.tensor('A',2,3)
B = ff.tensor('B',3,3)
#tsrex = A[[*i, *j]] * B[i,j] # Expect -> 6 x 9, get: 6 x 9 
tsrex = A[[*i, ~j]] * B[i,j] # Expect -> 6 x 3, get: 3 x 6 !!
#tsrex = A[[*i,  j]] * B[i,j] # Expect-> 6, get: 6 
#tsrex = A[[i,   j]] * B[j,k] # Expect -> 2 x 3, get: 2 x 3
#tsrex = A[[i,  ~j]] * B[j,k] # Expect -> 2 x 3 x 3, get: 2 x 3 x 3
#tsrex = A[[i,  *j]] * B[j,k] # Expect -> 2 x 9 x 3, get: 2 x 9 x 3
tsrex = tsrex | _my_index  | _my_init | _my_ein
out = tsrex | _my_eval
print(out.shape)
```

### DIFF
--- a/funfact/lang/interpreter/_einop.py
+++ b/funfact/lang/interpreter/_einop.py
@@ -48,12 +48,11 @@ def _einop(spec: str, lhs, rhs, reduction: str, pairwise: str):
     pairwise: str
         Name of the pairwise operator
     '''
-
     # parse input spec string
-    lhs_spec, rhs_spec, res_spec, kron_spec = re.split(r'\W+', spec)
+    lhs_spec, rhs_spec, out_spec, kron_spec = re.split(r'\W+', spec)
     lhs_spec = list(lhs_spec)
     rhs_spec = list(rhs_spec)
-    res_spec = list(res_spec)
+    out_spec = list(out_spec)
     kron_spec = list(kron_spec)
 
     # reorder lhs and rhs in alphabetical order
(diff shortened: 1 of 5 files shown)


yhtang:
Could we treat the `kron` indices as a subcategory of `keep` indices? Would this fix the shape error that we see in e.g.
```
tsrex = A[[*i, ~j]] * B[i,j] # Expect -> 6 x 3, get: 3 x 6 !!
```
?

funfact/lang/interpreter/_einop.py
```diff
@@ -85,15 +86,30 @@ def _einop(spec: str, lhs, rhs, reduction: str, pairwise: str):
...
+    tile_rhs = []
+    for i, c in enumerate(indices_all):
+        if c in kron_spec:
+            tile_lhs.append(shape_rhs[i])
+            tile_rhs.append(shape_lhs[i])
+        else:
+            tile_lhs.append(1)
+            tile_rhs.append(1)
+    lhs = np.tile(lhs, tuple(tile_lhs))
+    rhs = np.tile(rhs, tuple(tile_rhs))
```
yhtang, line 106: I really like the `tile` approach, which saves the creation of the 1 vector. However, I guess this only works for the RHS? e.g. for LHS, we want to get `[a, b, c] -> [a, a, a, b, b, b, c, c, c]`, while tiling gives `[a, b, c, a, b, c, a, b, c]`.

campsd:
> Could we treat the `kron` indices as a subcategory of `keep` indices? Would this fix the shape error that we see in e.g.
> 
> ```
> tsrex = A[[*i, ~j]] * B[i,j] # Expect -> 6 x 3, get: 3 x 6 !!
> ```
> 
> ?

Yes, I believe so, changed this in the latest commit and it seems to work correctly. Also updated the shape analyzer now.

funfact/lang/interpreter/_einop.py
campsd: That's right! The current implementation is wrong. I'll think a bit if it's best to keep the tiling for the RHS and use the 1 vector for the LHS, or maybe change both to the 1 vector.

yhtang:
> > Could we treat the `kron` indices as a subcategory of `keep` indices? Would this fix the shape error that we see in e.g.
> > ```
> > tsrex = A[[*i, ~j]] * B[i,j] # Expect -> 6 x 3, get: 3 x 6 !!
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ?
> 
> Yes, I believe so, changed this in the latest commit and it seems to work correctly. Also updated the shape analyzer now.

Cool!

funfact/lang/interpreter/_einop.py
campsd: It should be fixed now. New implementation uses broadcasting for both Kronecker and free indices and does a reshape at the end to get the Kronecker ordering.

yhtang (approved):

Issue #54, opened 2021-12-02
Expand the set of optimizers and loss functions


campsd:
We currently have the `Adam` optimizer and the `MSE (L2)` loss function.

We should add a few more options to the optimization library:

*Optimizers*

- RMSProp
- AdaGrad
- ... 

*Loss functions*
- L1 loss
- KL divergence
- ...

Additionally, we should allow the user to add a **penalty term** to the loss function to enforce (unitary) constraints.

yhtang:
We might need to wrap up this issue as soon as possible, since:
1. it is blocking the demo & documentation of `factorize`.
2. it might fall out of sync with all the other changes, in particular the backend PR #75.

I thought we already have most of the parts, right? :smiling_imp:

PR #57, opened 2021-12-03
[WIP] Optimizers

yhtang:
This is a subtask from #54.

Given that we will implement just a few gradient descent optimizers, it is kind of an overkill to structure it as a Python package (aka folder). I've reduced it to a single module (aka file) and we can just put in 3 or 4 of the algorithms there.


yhtang:
All changes in this PR are now included in #67.

PR #62, opened 2021-12-06
Devectorizer

campsd:
A devectorizer can be used to get one instance of a vectorized factorization model and return it as a new model.

*Code snippet:*
```Python
A = ff.tensor('A',2,3)
B = ff.tensor('B',3,4)
i,j, k = ff.indices('i, j, k')
tsrex = A[i,j] * B[j, k]
fac = Factorization(tsrex,nvec=5)
fac_scalar = Factorization.devectorize(fac, 1)
print(fac()[...,1])
print(fac_scalar())
```

Both should return the same output.

### DIFF
--- a/funfact/model/_factorization.py
+++ b/funfact/model/_factorization.py
@@ -59,15 +59,14 @@ class Factorization:
     def shape(self):
         return self._tsrex.root.shape
 
-    @classmethod
-    def devectorize(cls, factorization, instance: int):
-        '''Devectorize a factorizaition and keep a single slice.'''
-        if instance >= factorization.nvec:
+    def devectorize(self, instance: int):
+        '''Devectorize a factorizaition and keep a single instance.'''
+        if instance >= self.nvec:
             raise IndexError(
-                f'Index {instance} out of range (nvec: {factorization.nvec})'
+                f'Index {instance} out of range (nvec: {self.nvec})'
             )
-        tsrex = factorization.tsrex | Devectorizer(instance)
-        return cls(tsrex, initialize=False)
+        tsrex = self.tsrex | Devectorizer(instance)
+        return type(self)(tsrex, initialize=False)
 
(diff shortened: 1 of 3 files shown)


funfact/model/_factorization.py
```diff
@@ -44,15 +45,30 @@ def __init__(self, tsrex, initialize=True, nvec=1):
...
 
+    @property
+    def nvec(self):
+        return self._nvec
+
     @property
     def shape(self):
         return self._tsrex.root.shape
 
+    @classmethod
```
yhtang, line 62: Why do we need to make it a `classmethod`?

funfact/model/_factorization.py
```diff
@@ -44,15 +45,30 @@ def __init__(self, tsrex, initialize=True, nvec=1):
...
+    def devectorize(cls, factorization, slice: int):
```
yhtang, line 63: I feel that `i` or `instance` might be a better name than `slice`, since it is a single integer.

yhtang (changes requested):

funfact/model/_factorization.py
campsd: I was thinking of it as a constructor-like method that returns a new instance of the class, which is why this seemed appropriate.

funfact/model/_factorization.py
campsd: Renamed to `instance` in latest commit.

funfact/model/_factorization.py
yhtang: We can create an instance by `type(self)(tsrex, ...)`, so there is no need to make it a classmethod just for the purpose of obtaining `cls`. Besides, it could be tedious if users have to do
```
fac.devectorize(fac, i)
```
where `fac` is repeated twice.

yhtang (approved):

Issue #65, opened 2021-12-08
[REQ] Add support for Ellipsis to indexing function of factorization model


campsd:
We can currently evaluate a factorization model elementwise or slicewise, i.e.:

```
fac[0, 1, 2]
fac[1:2, :, 3]
```

are supported for factorization models. It would be useful to add support for `Ellipsis` indexing similar to numpy:

```
fac[0, ...]
fac[..., 1]
fac[1:2, ..., 3]
```

This is particularly useful for accessing instances of a vectorized model.

PR #66, opened 2021-12-08
Ellipsis indexing for factorization models + basic bounds checking

campsd:
Implements #65

### DIFF
--- a/funfact/model/_factorization.py
+++ b/funfact/model/_factorization.py
@@ -57,7 +57,11 @@ class Factorization:
 
     @property
     def shape(self):
-        return self._tsrex.root.shape
+        return self._tsrex.shape
+
+    @property
+    def ndim(self):
+        return self._tsrex.ndim
 
     def devectorize(self, instance: int):
         '''Devectorize a factorizaition and keep a single instance.'''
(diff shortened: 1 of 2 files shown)


funfact/model/_factorization.py
```diff
@@ -76,15 +80,52 @@ def forward(self):
...
-        new_idx = []
-        for i in idx:
+
+        # Generate full index list
+        full_idx = []
+        ellipsis_i = None
+        for i in key:
             if isinstance(i, int):
-                new_idx.append(slice(i, i+1))
+                if i != -1:
```
yhtang, line 91: Could `i` be a negative number other than -1?

funfact/model/_factorization.py
```diff
@@ -76,15 +80,52 @@ def forward(self):
...
+                full_idx.append(i)
+            elif i is Ellipsis:
+                ellipsis_i = key.index(...)
             else:
-                new_idx.append(i)
-        _index_slicer = SlicingPropagator(new_idx)
+                raise IndexError(
+                    f'Unrecognized index {i} of type {type(i)}'
+                )
+        if ellipsis_i is not None:
```
yhtang, line 103: Could this handle the case where `...` appears at the end or middle of the indices?

funfact/model/_factorization.py
campsd: I think anything between `-size_of_dimension` and `size_of_dimension-1` could make sense?

funfact/model/_factorization.py
campsd: Yes, normally it should because of the `.insert`. It is assumed however that there is at most one `...`. I think same assumption as numpy makes, otherwise you can't infer the sizes.

yhtang (approved):

PR #68, opened 2021-12-09
Use `>>` instead of `.T` to denote transposition.

### LINKED ISSUE #60: Ambiguous semantics of the `.T` transposition syntax
**Describe the bug**
Potnetial flaw in the `.T[...]`  transposition syntax:

When people write `$A^T_{ij}$`, they mean the (i, j)-th element of the transposition of A. Therefore, the Einstein notation `$A_{ij} \cdot A^T_{ij}$` means the inner product between A and its transposition.

In Funfact, however, if A is a tsrex with live indices `i, j`, then `A.T[i. j]` does nothing --- it merely means to reorder the live indices of A from `i, j` to `i, j`, so nothing happens.

This is because `.T` means **tranpose according to the following []** instead of directly creating a transposed tensor where the dimensions are reversed.

This could catch a user off-guard if they are trying to do a literal translation of some formula on paper to code using FunFact.

**On the other hand, with the Einstein notation, there is no need to use `.T` to obtain the transposition of a tensor for computation --- simply adjusting the order of the indices will do.** The only purpose of `.T` is to actually transpose the tensor in memory for storage purposes.

**To Reproduce**
```
A = ff.tensor('A', 3, 3)
tsrex = A * A.T[j, k]  # expects to get $A \cdot A^T$, won't work
tsrex = A[i, j] * A[k, j]  # this works and computes A * A^T
```

**Discussion**

Is there a need to adjust this?

yhtang:
Closes #60. This avoids the confusion that may arise as discussed in the issue.

### DIFF
--- a/funfact/lang/_tsrex.py
+++ b/funfact/lang/_tsrex.py
@@ -205,8 +205,10 @@ class IndexRenamingMixin:
 class TranspositionMixin:
     '''transpose the axes by permuting the live indices into target indices.'''
     def __rshift__(self, indices):
-        return TsrEx(P.tran(self.root,
-                     P.indices(tuple([i.root for i in as_tuple(indices)]))))
+        return TsrEx(P.tran(
+            self.root,
+            P.indices(tuple([i.root for i in as_tuple(indices)]))
+        ))
 
 
 class TsrEx(_BaseEx, ArithmeticMixin, IndexRenamingMixin, TranspositionMixin):
--- a/funfact/lang/_tsrex.py
+++ b/funfact/lang/_tsrex.py
@@ -204,16 +204,9 @@ class IndexRenamingMixin:
 
 class TranspositionMixin:
     '''transpose the axes by permuting the live indices into target indices.'''
-    @property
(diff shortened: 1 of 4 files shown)


funfact/lang/_tsrex.py
```diff
@@ -244,6 +237,7 @@ def __getitem__(self, indices):
 
 
 class EinopEx(TsrEx):
+    # override the `>>` behavior from TranspositionMixin
```
campsd, line 240: Do we need `>>` in both places? Is the value of the `TranspositionMixin` class to be able to permute a single tensor instead of an expression?

campsd (changes requested):
Best to also update the test cases that use `.T` notation

funfact/lang/_tsrex.py
yhtang: We do need both `__rshift__`, because the default one creates a new `tran` node, while the one here only modified the `outidx` field of the `ein` node.

campsd (approved):

PR #72, opened 2021-12-10
Bugfix/factorization view and einop

campsd:
I fixed two bugs while working to fix one of the two:

- `einop` didn't work correctly if one of the tensors is initialized from data and vectorized by adding a singleton dimension
- `view` method in factorization now also works for vectorized tensors from data with a singleton dimension

### DIFF
--- a/funfact/model/_factorization.py
+++ b/funfact/model/_factorization.py
@@ -70,7 +70,13 @@ class Factorization:
                 f'Index {instance} out of range (nvec: {self.nvec})'
             )
         fac = type(self)(self._otsrex, initialize=False)
-        fac.factors = [f[..., instance] for f in self.factors]
+        instance_factors = []
+        for f in self.factors:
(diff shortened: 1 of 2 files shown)


funfact/model/_factorization.py
```diff
@@ -70,7 +70,13 @@ def view(self, instance: int):
...
             )
         fac = type(self)(self._otsrex, initialize=False)
-        fac.factors = [f[..., instance] for f in self.factors]
+        instance_factors = []
+        for f in self.factors:
+            if f.shape[-1] == 1:
+                instance_factors.append(f[..., 0])
+            else:
+                instance_factors.append(f[..., instance])
+        fac.factors = instance_factors
```
campsd, line 79: `fac.view(0)` probably still fails if called on a factorization that is not vectorized. Could be checked on input?

PR #73, opened 2021-12-11
Allow definition of zeroth-order tensors, i.e. scalars. Closes #69.

### LINKED ISSUE #69: Zero dimensional tensors
For constant optimizable offsets in a tensor expression, we should add the support for zero-dimensional tensors.

An example use case and suggested syntax is given in the following RBF expression:
```Python
# RBF approximation, one instance
m, n = 24, 36
r = 8
u = ff.tensor('u', m, r)
v = ff.tensor('v', n, r)
a = ff.tensor('a', r)
b = ff.tensor('b')
tsrex = a[k] * ff.exp(-(u[i, ~k] - v[j, ~k])**2) + b[:] # or b[...]
```

yhtang:
Since a zeroth-order tensor, i.e. a scalar, is one that does not carry any indices, it should be semantically straightforward to allow a zeroth-order tensor to be defined as one whose shape is empty (not even as `[1]`!) and which can be indexed by `[None]` (since there _is_ no indices for it). It turns out the rest of FunFact already works with zeroth-order tensors out of the box.

Example
```
a = ff.tensor('a')
b = ff.tensor()
print(a.ndim)  # expects 0
print(a.shape)  # expects (,)
print(b.live_indices)  # expects []

fac = Factorization(a[[]])
print(fac())  # obtains a single-element vector
vfac = Factorization(b[None], nvec=4)
print(vfac())  # obtains a 4-element vector
```

### DIFF
--- a/funfact/util/typing.py
+++ b/funfact/util/typing.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
+import numpy as np
 
 
 def _is_tensor(x):
-    return hasattr(x, "__len__")
+    return isinstance(x, np.ndarray)
--- a/funfact/lang/_tsrex.py
+++ b/funfact/lang/_tsrex.py
@@ -232,7 +232,7 @@ class TensorEx(_BaseEx):
             P.index_notation(
                 self.root,
                 P.indices(
-                    tuple([i.root for i in as_tuple(indices)])
+                    tuple([i.root for i in as_tuple(indices or [])])
                 )
             )
         )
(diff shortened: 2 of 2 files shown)


funfact/util/typing.py
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
+import numpy as np
 
 
 def _is_tensor(x):
-    return hasattr(x, "__len__")
+    return isinstance(x, np.ndarray)
```
campsd, line 7: Does this work with the JAX `DeviceArray` type?

campsd (approved):

Issue #70, opened 2021-12-09
Vectorization in Factorization model


campsd:
Related to #54, #57  

Add full support for vectorization to the factorization model.

Some considerations:

- It probably makes sense to consider _every_ factorization as vectorized even if `nvec=1`
- Bookkeeping of all the best solutions can happen in a copy
- Total loss = sum over vectorized loss function

yhtang:
We discussed earlier today about automatically inferring the vectorization dimensions by shape comparison.

Issue #55, opened 2021-12-02
Refactor package for v1.0 release


campsd:
There are a few refactoring issues to resolve for the initial release.

Move all files to appropriate (new) modules:

- `_einop` should be moved from the `lang>interpreters` to `algorithm` module
- look into reorganizing the `optim` module
- `nn` neural network module
- rethink the interface to `Factorization` and the `factorize` function: 1) `fit` method for factorization, 2) add `vectorize` option, 3) `tol` convergence checking, ...
- ...

Expose all required functionality to the enduser: `Factorization`, `factorize`, `neural network`, ...

yhtang:
We can make sure `Factorization` is accessible at the top level, and then this should eliminate the need for #32.

yhtang:
Closing it now as its scope has been well exceeded.

PR #77, opened 2021-12-15
Loss class

campsd:
Class for loss functions.

Part of #54, #55 and required for #70.

### DIFF
--- a/funfact/loss.py
+++ b/funfact/loss.py
@@ -1,0 +1,49 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+import jax.numpy as np
+
+
+class Loss:
+
+    def __init__(self, type='mse'):
+        if type == 'mse':
+            def _loss(model, target):
+                return np.square(np.subtract(model, target))
+        elif type == 'l1':
+            def _loss(model, target):
+                return np.abs(np.subtract(model, target))
+        elif type == 'kl_divergence':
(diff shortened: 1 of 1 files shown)


funfact/loss.py
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+import jax.numpy as np
```
yhtang, line 3: Now we can replace line with the backend import. 😎

funfact/loss.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+class Loss:
```
yhtang, line 6: I think it'll be better to actually split this to three classes, one for each loss function. This one currently looks like a factory method. Do you have some concern regarding separating them?

yhtang (changes requested):
If you are worried about duplicating code, e.g.  for the data dim checks, that can be done by factoring out the function as a private method (aka `_` prefixed).

funfact/loss.py
campsd: Done.

funfact/loss.py
campsd: I have split it into 3 classes. If there are loss functions that differ from the default behavior, they can override the `__call__` function from the base class.

yhtang (approved):

PR #86, opened 2021-12-19
Full functionality for using the PyTorch backend. Fixes #76. Fixes #98.

### LINKED ISSUE #76: [BUG] `einop` does not work with the PyTorch backend
**Describe the bug**
Our current `einop` implementation depends on `order='F'` of `numpy.transpose` or `jax.numpy.transpose`, which PyTorch does not implement. Simply removing the order specifier leads to incorrect results.

**To Reproduce**
```
ff.use('torch')
u = ff.tensor(np.arange(3))
v = ff.tensor(np.ones(4))
i = ff.index()
fac = Factorization(u[[*i]] * v[[*i]])
fac()
```

**Expected behavior**
`[0., 0., 0., 0., 1., 1., 1., 1., 2., 2., 2., 2.]`

**Additional context**
This StackOverflow [answer](https://stackoverflow.com/a/63964246) provides a solution that is potentially slow. Also, I tried to directly cast the solution and ended up getting a single number instead of 12 elements.

### LINKED ISSUE #98: Numpy compatibility of torch backend

campsd:
This fixes #76, I haven't checked the performance of the  `reshape(..., order='F')` implementation.

The `torch.sum` method is overridden to have the same behavior as `numpy` if axis is an empty tuple. The same will have to be don for other reduction operations: `torch.prod`, `torch.max`, `torch.min`, ...

### DIFF
--- a/funfact/algorithm.py
+++ b/funfact/algorithm.py
@@ -62,6 +62,8 @@ def factorize(
             that represents all the solutions.
     '''
 
+    target = ab.tensor(target)
+
     @ab.autograd_decorator
     class _Factorization(Factorization, ab.AutoGradMixin):
         pass
@@ -92,7 +94,6 @@ def factorize(
 
     tsrex_vec = vectorize(tsrex, nvec)
     opt_fac = _Factorization.from_tsrex(tsrex_vec)
-    best_fac = Factorization.from_tsrex(tsrex_vec)
 
     try:
         opt = optimizer(opt_fac.factors, lr=lr, **kwargs)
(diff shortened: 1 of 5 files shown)


yhtang:
This looks pretty good to me. There will be some performance hit for the double transposition in reshape but let's worry about that later.

yhtang:
I do see that the unit tests fail because we don't have torch listed as a dependency yet. Maybe for the time being, we can test the correctness using the NumPy backend?

campsd:
This PR adds full functionality for using the PyTorch backend with FunFact.

Furthmore, it solves:

* #98 by providing `to_numpy` method in the backend and by converting the target data in `factorize` to native tensor type
* Bugfix of #76 

The only thing missing is the `jit.trace` compatibility with the `_Factorization` class.

funfact/algorithm.py
```diff
@@ -101,52 +102,53 @@ class _Factorization(Factorization, ab.AutoGradMixin):
...
-            else:
-                if stop_by is not None:
-                    raise RuntimeError(
-                        f'Invalid argument value for stop_by: {stop_by}'
-                    )
+        _, grad = loss_and_grad(opt_fac, target)
+        with ab.no_grad():
+            opt.step(grad)
+
+            if step % round(max_steps/20) == 0:
```
yhtang, line 119: It might feel a bit arbitrary if we are updating the best solution by every `round(max_steps/20)`. How about making it an argument e.g. `checkpoint_freq` that defaults to 1?

funfact/algorithm.py
```diff
@@ -101,52 +102,53 @@ class _Factorization(Factorization, ab.AutoGradMixin):
...
+                # update best factorization
+                curr_loss = loss(opt_fac(), target, sum_vec=False)
+                new_best = []
+                for b, o in zip(best_factors, opt_fac.factors):
+                    for i, l in enumerate(zip(curr_loss, best_loss)):
```
yhtang, line 124: I believe if we flip line 123 and 124, the loop over the current line 124 can be vectorized using np array operations (c.f. my old implementation). :grin:

yhtang (changes requested):
Great job! There are two places that I think we could improve a bit. Let me know what you think.

funfact/algorithm.py
campsd: I have added this argument but set the default value to 50. Checking every single iteration by default seemed a bit much to me.

funfact/algorithm.py
campsd: I have flipped the order and made an improvement that got rid of the unnecessary copy to `new_best` so that's a good improvement but I didn't find a proper way to get completely rid of the second loop with array operations or list comprehensions. Let me know what you think of this version. :)

yhtang (approved):

PR #88, opened 2021-12-19
[WIP] Interpreter redesign

yhtang:
This is an effort to reduce the LOCs for the interpreter module to facilitate testing. However, it seems that the reduction is small and may not justify the additional complexity that is introduced.

### DIFF
--- a/funfact/lang/interpreter/_ascii.py
+++ b/funfact/lang/interpreter/_ascii.py
@@ -9,8 +9,8 @@ class ASCIIRenderer(TranscribeInterpreter):
     as_payload = TranscribeInterpreter.as_payload('ascii')
 
     @as_payload
-    def noop(self, **kwargs):
-        return f'<unknown primitive {kwargs}>'
+    def noop(self):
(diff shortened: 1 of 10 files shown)


yhtang:
I'm actually now in favor of abandoning this PR, but not 100% certain... :joy:

campsd:
The idea is OK in my opinion. It reduces the complexity for some of the simple interpreters that act on only a few type of nodes.

yhtang:
> The idea is OK in my opinion. It reduces the complexity for some of the simple interpreters that act on only a few type of nodes.

Thanks for the comments!

I just took a thought experiment, pretending that I am new to the project and was trying to understand one of the interpreters. It turns out figuring out the 'missing' primitives, i.e. those that default to a no-op, causes a bigger mental burden. I will essentially have to view the concrete interpreter and the base interpreter side-by-side to achieve this.

As such, I think I'll close this PR without merging. However, I think we can keep it for a while since I might have discovered a few potential :lady_beetle: along the way.

PR #89, opened 2021-12-19
iterable factorization model for torch

campsd:
Small PR to make the factorization model iterable as required for autograd in torch


yhtang:
Could we try to make the mini-model in the Torch backend iterable instead? My concern here is that `Factorization.factors` already supports iteration, hence providing iterability to the factorization object itself could be redundant and confusing. What do you think?

campsd:
Yeah, makes sense. I'll move the `__iter__` method to the `AutoGradMixin` for `torch`

Issue #38, opened 2021-11-22
Optimizable properties for terminal nodes in a tensor expression


campsd:
Currently all `tensors` in a tensor expression are updated in the gradient descent step of a factorization model. We should implement a feature such that the user can decide which of the `tensors` (and maybe also `LiteralValues`?)  are actually optimized.

Potential syntax:
```
A = ff.tensor('A', 2, 3, optimizable=True)
offset = ff.literal('o', optimizable=True)
...
```

yhtang:
In addition to explicitly specifying the `optimizable` attribute, we can also assume by default that an abstract tensor instantiated with only shape information will be optimizable (since they will be random initialized), while one made from a concrete tensor will be constant, e.g.
```
A = ff.tensor('A', 2, 3, 4)  # A defaults to optimizable
B = ff.tensor('B', np.eye(3))  # B defaults to non-optimizable.
```

yhtang:
Also makes sense to provide `.set_optimizable(True/False)` methods for the terminal nodes.

campsd:
This is works nicely with `torch` where optimizable tensors have to be created by setting `requires_grad=True`

yhtang:
> This is works nicely with `torch` where optimizable tensors have to be created by setting `requires_grad=True`

Yep. What is the JAX equivalence of it?

campsd:
I think not returning the non-optimizable leafs in `Factorize.factors` so JAX can't trace them? I'm working on implementing this feature now.

yhtang:
> I think not returning the non-optimizable leafs in `Factorization.factors` so JAX can't trace them? I'm working on implementing this feature now.

I see. Makes sense. Thanks!

yhtang:
Implemented by #97.

PR #90, opened 2021-12-20
Towards #38

campsd:
This draft PR is meant to implement #38 

I see two remaining issues with current version:

- `set_optimizable` function is part of `AbstractTensor`. This has the following effect:
```Python
a = ff.tensor(2, 3, 4)
a.set_optimizable(False) # This doesnt work
a.root.abstract.set_optimizable(False) # This does work but is too complicated
```

- If `set_optimizable` is changed _after_ the leaf initializer is run, it will work for a JAX backend (the `dfs_filter` should be sufficient), but it won't work properly for torch as we need to change the `requires_grad` flag of the tensor. Maybe we can check in the `set_optimizable` function if: (1) the backend is torch, and (2) the tensor is already initialized and change the `requires_grad` flag.

What do you think @yhtang ?

### DIFF
--- a/funfact/backend/_jax.py
+++ b/funfact/backend/_jax.py
@@ -18,7 +18,7 @@ class JAXBackend(metaclass=BackendMeta):
     tensor_t = (jnp.ndarray, np.ndarray)
 
     @classmethod
-    def tensor(cls, array, **kwargs):
+    def tensor(cls, array, optimizable, **kwargs):
         return jnp.asarray(array, **kwargs)
 
     @classmethod
@@ -26,7 +26,7 @@ class JAXBackend(metaclass=BackendMeta):
         cls._key = jrn.PRNGKey(key)
 
     @classmethod
-    def normal(cls, mean, std, *shape, dtype=jnp.float32):
+    def normal(cls, mean, std, *shape, optimizable, dtype=jnp.float32):
         cls._key, subkey = jrn.split(cls._key)
         return mean + std * jrn.normal(subkey, shape, dtype)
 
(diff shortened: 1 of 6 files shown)


yhtang:
> a.set_optimizable(False) # This doesnt work

Having `set_*` methods on the AST nodes and/or the abstract tensors would violate the immutability of our functional programming model. I think it is better to specify whether a tensor is optimizable at initialization, e.g., with
``` py
a = ff.tensor('a', 3, 2, 4, mutable=True)
```
A side issue here is how to name this option. I think `optimizable` is great except for being a bit too long. `mutable` is shorter but perhaps not as intuitive.

Further, we could revive the eagerness context work and repurpose that to adjust the optimizability of tensors. It can be either a global on/off switch, or be specialized to act on select tensors. Examples would be:
``` py
with ff.mutable/optimizable(True/False):  # global on/off switch
    ...

with ff.mutable/optimizable(a=True, b=False):  # a, b are symbols of tensors
    ...
```

campsd:
Ok, makes sense to only allow it at initialization.

About the name: `optimizable` is the same length as `initializer`, the other kwarg for the `tensor` method. Maybe we can reduce both to `opt` (or `optim`?) and `init`?

Also, wouldn't the context manager also violate the immutability in the same way as a `set_*` method does?

yhtang:
Documenting what we discussed over Zoom just for the record:

> About the name: `optimizable` is the same length as `initializer`, the other kwarg for the `tensor` method. Maybe we can reduce both to `opt` (or `optim`?) and `init`?

Turns out `initializer` is a well-established convention in use since Keras, so we'll probably gonna keep it. For `optimizable`/`mutable`, I am now in favor of `optimizable` since its meaning is more direct. :grin: 

> Also, wouldn't the context manager also violate the immutability in the same way as a `set_*` method does?

This should be fine since contexts are **reversible**.

PR #94, opened 2021-12-21
Separated the vectorization mechanism from the Factorization class

yhtang:
> Yu-Hang  1 hour ago
> I am thinking about making the Factorization class agnostic of vectorization. This should remove many complications surround what if nvec=1.
> 
> Yu-Hang  1 hour ago
> The factorize method will carry out vectorization regardless, though.
> 
> Yu-Hang  1 hour ago
> However, vectorization will not be done at the initialization of a Factorization object. It will be done by a pair of standalone vectorize/view/devectorize methods, kind of like what JAX does.
> 
> Daan Camps  43 minutes ago
> That sounds like a neat idea. Probably the vectorize and view methods are sufficient? A devectorize is not really required if we have a view
>

### DIFF
--- a/funfact/algorithm.py
+++ b/funfact/algorithm.py
@@ -1,14 +1,17 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+import numpy as np
+import tqdm
 import funfact.optim
 import funfact.loss
 from funfact import Factorization
 from funfact.backend import active_backend as ab
-import tqdm
-import numpy as np
+from funfact.vectorization import vectorize, view
 
 
 def factorize(
     tsrex, target, lr=0.1, tol=1e-6, max_steps=10000, optimizer='Adam',
-    loss='mse_loss', nvec=1, stop_by='best', **kwargs
+    loss='mse_loss', nvec=1, stop_by='first', returns='best', **kwargs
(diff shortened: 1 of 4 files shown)


funfact/algorithm.py
```diff
@@ -102,22 +132,28 @@ def _loss(fac, target):
                 new_best.append(b)
             best_fac.factors = new_best
 
-            if stop_by == 'best':
+            if stop_by == 'first':
                 if np.sum(converged) > 1:
```
campsd, line 136: I just realize that we can stop already if `np.sum(converged) > 0` in this case

funfact/algorithm.py
```diff
@@ -102,22 +132,28 @@ def _loss(fac, target):
...
                     pbar.update(max_steps - step)
                     break
             elif isinstance(stop_by, int):
                 if np.sum(converged) > stop_by:
```
campsd, line 140: Same here `np.sum(converged) > stop_by - 1`

campsd (approved):
Looks good to me.

funfact/algorithm.py
yhtang: Good point. I'll change that.

PR #97, opened 2021-12-21
Optimizable functionality

campsd:
This implements #38.

Four core features:

1. At creation of a `tensor`, a flag `optimizable`  can be set by the user or determined by the type of input. By default the following behavior will happen:
```Python
a = ff.tensor('a', 2, 3) # optimizable will be True as randomly initialized
b = ff.tensor(2, 3) # same as above
c = ff.tensor('c', 2, 3, initializer=np.randn) # same as above
d = ff.tensor('2, 3, initializer=np.randn) # same as above
e = ff.tensor('e', np.ones(2,3)) # optimizable will be False as concrete data is given
f = ff.tensor(np.ones(2,3)) # same as above
```

2. The initializers of the different backends are adjusted to the `optimizable` flag. Mostly relevant for PyTorch.

3. The `factors` property of the `Factorization` model only returns the `optimizable` leaf tensors. The `all_factors` method is added to return all leaf tensors including non-optimizable

4. A context manager is added to override the `optimizable` setting. Mainly relevant for PyTorch backend:
```Python
with ff.set_optimizable(False):
    ff.optimizable.is_optimizable() # False

ff.optimizable.is_optimizable() # True
```


funfact/__init__.py
```diff
@@ -6,6 +6,8 @@
 from .lang._math import *  # noqa: F401, F403
 from .model import Factorization
 from .algorithm import factorize
+from.optimizable import set_optimizable
```
yhtang, line 9: Looks like a space is missing here :grin:

funfact/backend/_torch.py
```diff
@@ -14,16 +14,20 @@ class PyTorchBackend(metaclass=BackendMeta):
...
     def seed(cls, key):
         cls._gen.manual_seed(key)
 
     @classmethod
-    def normal(cls, mean, std, *shape, dtype=torch.float32):
-        return torch.normal(mean, std, shape, dtype=dtype, generator=cls._gen)
+    def normal(cls, mean, std, *shape, optimizable, dtype=torch.float32):
+        data = torch.normal(mean, std, shape, dtype=dtype, generator=cls._gen)
+        if optimizable:
+            return data.clone().detach().requires_grad_(True)
```
yhtang, line 28: This could cause some unnecessary overhead due to `clone`. Can we directly override the `requires_grad` property?

funfact/lang/_terminal.py
```diff
@@ -152,7 +152,7 @@ class TensorSymbol(Symbol):
         _anon_registry = {}
         _anon_registry_lock = multiprocessing.Lock()
 
-    def __init__(self, *size, symbol=None, initializer=None):
+    def __init__(self, *size, symbol=None, initializer=None, optimizable=None):
```
yhtang, line 155: Since `optimizable` is a boolean, the default value should be either `True` or `False`.

funfact/lang/_tsrex.py
```diff
@@ -325,6 +325,10 @@ def tensor(*spec, initializer=None):
         initializer (callable):
             Initialization distribution
 
+        optimizable (boolean):
```
yhtang, line 328: Based on the PR description, the default value is dependent on `spec`, right?

yhtang: Some examples in the docstring should help to clarify this behavior.

funfact/model/_factorization.py
```diff
@@ -79,16 +79,46 @@ def factors(self):
...
+            dfs_filter(lambda n: n.name == 'tensor' and
+                       n.abstract.optimizable, self.tsrex.root)
         ):
             n.data = tensors[i]
 
+    @property
+    def all_factors(self):
+        '''A flattened list of all factors in the model.
+
+        Examples:
```
yhtang, line 98: Could you please update the example in the docstring to reflect the difference between `factors` and `all_factors`?

funfact/optimizable.py
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env python
```
yhtang, line 1: Perhaps a better name for this file would be `context.py`?

funfact/optimizable.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def __init__(self, mode: bool) -> None:
+        global _all_optimizable
+        self._prev_mode = _all_optimizable
+        _all_optimizable = mode
+        if 'torch' in ab.__repr__().lower():
+            ab.set_grad_enabled(mode)
+
+    def __enter__(self) -> None:
+        pass
```
yhtang, line 17: Having an empty `__enter__` is a bit suspicious to me.

yhtang:
Nice job! I added some nit-picks. Let me know what you think.

funfact/backend/_torch.py
campsd: This is a solution I found somewhere on a forum. We don't want to set the `requires_grad` property before the random data is generated as then the tracing could include the rng method.

This is a relevant thread: https://stackoverflow.com/questions/51751231/pytorch-effect-of-normal-initialization-on-gradients

Maybe a solution could be to create an `as_tensor` method that can create torch `tensor`s from numpy array and use the numpy rng in the `normal` function for the torch backend.

Something like:
```
def normal(cls, mean, std, *shape, optimizable, dtype=torch.float32):
    data = np_backend.normal(cls, mean, std, *shape)
    return as_tensor(data, optimizable=optimizable, dtype=dtype)
```

What do you think?

funfact/optimizable.py
campsd: This is actually quite similar to the PyTorch implementation: https://pytorch.org/docs/stable/_modules/torch/autograd/grad_mode.html#set_grad_enabled

Do you think it's better to move the `__init__` code to `__enter__`?

funfact/backend/_torch.py
yhtang: Yeah, honestly I also recall reading this quirk with PyTorch a long while ago. The numpy solution does not seem very attractive either :rofl: I guess we can just settle with what it is for now.

funfact/optimizable.py
yhtang: Given that this context manager is relatively lightweight, can we simply use [contextlib](https://docs.python.org/3/library/contextlib.html) to create it?

yhtang: I mean the `@contextlib.contextmanager` decorator :smile:

PR #108, opened 2021-12-24
Allow hybrid indexed/indexless tensor arithmetics

yhtang:
This PR enables mixed tensor algebra expressions. Essentially, it makes our tensor expression programming model a **superset** of what NumPy implements.

``` py
import funfact as ff
a = ff.tensor('a', 2, 4)
b = ff.tensor('b', 2, 4)
c = ff.tensor('c', 4, 10)
tsrex = a[i, j] * b[k, j]  # OK, Einstein notation, implied inner product, a @ b.T
tsrex = a * b  # OK, NumPy-style elemtwise operations
tsrex = a @ b.T  # OK, NumPy-style dot product
tsrex = (a + b)[i, j] * c[j, k]  # OK, first computes elementwise add between a and b, and then dot product with c
tsrex = b.T @ (a[i, j] * c[j, k])  # Will be OK after `@` implemented in the next PR.
```

### DIFF
--- a/funfact/lang/_ast.py
+++ b/funfact/lang/_ast.py
@@ -54,7 +54,9 @@ class Primitives:
 
     @primitive(precedence=1)
     def index_notation(indexless: _ASNode, indices: _ASNode):
-        '''indexing a raw tensor or indexless expression: expr[indices...]'''
+        '''indexing a raw tensor or indexless expression: expr[indices...].
+        To be interpreted either as tensor indexing or index renaming depending
+        on the type of the addressee.'''
 
     @primitive(precedence=2)
     def call(f: str, x: _ASNode):
@@ -73,8 +75,9 @@ class Primitives:
         '''indexless Kronecker product'''
 
     @primitive(precedence=None)
-    def elem(lhs: _ASNode, rhs: _ASNode, precedence: int, oper: str):
-        '''indexless elementwise operations'''
+    def binary(lhs: _ASNode, rhs: _ASNode, precedence: int, oper: str):
+        '''generic binary operations, to be interpreted as either elementwise
+        or Einstein based on index/indexless status.'''
 
     @primitive(precedence=None)
     def ein(
--- a/funfact/lang/_ast.py
+++ b/funfact/lang/_ast.py
@@ -90,18 +90,6 @@ class Primitives:
     def tran(src: _ASNode, indices: _ASNode):
         '''transposition/axis reordering'''
 
-    @classmethod
-    def as_primitive(cls, raw):
-        if isinstance(raw, _ASNode):
-            return raw
-        elif isinstance(raw, Real):
-            return cls.literal(value=LiteralValue(raw))
-        else:
-            raise TypeError(
-                f'Cannot use {raw} of type {type(raw)} in '
-                f'a tensor expression.'
-            )
-
     Tensorial = Union[
         index_notation, call, pow, neg, ein
     ]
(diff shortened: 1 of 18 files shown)


funfact/lang/_tsrex.py
```diff
@@ -167,105 +176,99 @@ def einspec(self):
...
 
-    def __getitem__(self, indices):
+    @as_tsrex
+    def __radd__(self, lhs):
+        return _binary(_as_node(lhs), _as_node(self), 6, 'add')
 
-        tsrex = self | IndexPropagator()
+    @as_tsrex
+    def __rsub__(self, lhs):
+        return _binary(_as_node(lhs), _as_node(self), 6, 'add')
```
campsd, line 237: Should this be `'sub'`?

funfact/lang/_tsrex.py
```diff
@@ -279,41 +282,59 @@ class TsrEx(_BaseEx, ArithmeticMixin, IndexRenamingMixin, TranspositionMixin):
...
-
     '''
-    pass
 
 
-class IndexEx(_BaseEx):
-    def __invert__(self):
-        '''Implements the `~i` syntax.'''
-        return IndexEx(dataclasses.replace(self.root, bound=True, kron=False))
+_dispatch = Dispatcher()
```
campsd, line 288: I don't fully understand the need for multiple dispatch here. There seems to be only one implementation of every function? Is it to raise an error if it is called with incorrect input?

funfact/lang/interpreter/_base.py
```diff
@@ -121,65 +131,90 @@ def wrapped_f(*args, **kwargs):
...
+
+        def _do_node():
+            payload = rule(**node.fields)
+            self.emplace(node, payload)
+
+        def _do_children():
+            for name, value in node.fields_fixed.items():
+                setattr(node, name, _deep_apply(self, value, node))
+
+        if self._traversal_order == self.TraversalOrder.PRE:
```
campsd, line 212: Great that you added both options in a clean way!

funfact/lang/_ast.py
```diff
@@ -68,6 +70,15 @@ def pow(base: _ASNode, exponent: _ASNode):
     def neg(x: _ASNode):
         '''elementwise negation'''
 
+    @primitive(precedence=5)
+    def kron(lhs: _ASNode, rhs: _ASNode):
+        '''indexless Kronecker product'''
```
campsd, line 75: What is the syntax for an indexless Kronecker product? I don't see this explained in the PR.

funfact/lang/_tsrex.py
yhtang: Definitely. Let me fix it.

funfact/lang/_tsrex.py
yhtang: You are right :smile: There are two purposes:
1. We will need a redirection anyway to handle `__op__` and `__rop__` in a single place.
2. Using multiple dispatch is a systematic way to catch certain types of syntax errors.

funfact/lang/interpreter/_base.py
yhtang: :sunglasses:

funfact/lang/_ast.py
yhtang: This is fully implemented with the `&` operator in a later PR #124. Honestly, I'm not completely satisfied with the use of `&`, because it has a very low operator precedence which is not consistent with the 'product' nature of the operation. I am not sure if there will be better options, though.

campsd (commented):
Looks great!

PR #121, opened 2022-01-01
Make exponentiation a special case of binary operations.

yhtang:
Also fixes a bug that unnecessarily excludes tensors of broadcastable shapes to act on each other in binary operations.

``` py
a = ff.tensor('a', 2, 3)
b = ff.tensor('b', 2, 3)
fac = ff.Factorization.from_tsrex(a**b)
print(fac())
print(fac.tsrex.asciitree)
fac.tsrex
```
Output:
```
[[      nan 0.6022982 1.4585489]
 [      nan       nan       nan]]
 binary: float_power 
 ├── tensor: a 
 ╰── tensor: b 
```

### DIFF
--- a/funfact/lang/_tsrex.py
+++ b/funfact/lang/_tsrex.py
@@ -234,7 +234,7 @@ class SyntaxOverloadMixin:
 
     @as_tsrex
     def __rsub__(self, lhs):
-        return _binary(_as_node(lhs), _as_node(self), 6, 'add')
+        return _binary(_as_node(lhs), _as_node(self), 6, 'subtract')
 
     @as_tsrex
     def __rmul__(self, lhs):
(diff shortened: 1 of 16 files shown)


funfact/lang/_tsrex.py
```diff
@@ -230,7 +230,7 @@ def __truediv__(self, rhs):
 
     @as_tsrex
     def __pow__(self, rhs):
-        return _pow(_as_node(self), _as_node(rhs))
+        return _binary(_as_node(self), _as_node(rhs), 3, 'float_power')
```
campsd, line 233: `float_power` seems a bit of a weird name to me. maybe just `power` is better? Similar to the name in NumPy.

campsd: Does the backend have to implement a `float_power` method for this to work?

yhtang: That's what I found out, `float_power` is the method that gets implemented in all three backends. Other than always promoting integers to floating points (which we almost always want), there is no side effect. :grin:

PR #124, opened 2022-01-01
Indexless Kronecker product using the `&` operator.

yhtang:
```
a = ff.tensor('a', 2, 3)
b = ff.tensor('b', 3, 1)
c = ff.tensor('c', 4, 3)
fac = ff.Factorization.from_tsrex(a & b & c)
print(fac().shape)
print(fac.tsrex.asciitree)
```
Output:
```
(24, 9)
 kron:  
 ├── kron:  
 │   ├── tensor: a 
 │   ╰── tensor: b 
 ╰── tensor: c 
```

### DIFF
--- a/funfact/__init__.py
+++ b/funfact/__init__.py
@@ -2,7 +2,7 @@
 # -*- coding: utf-8 -*-
 """Fun-Fact"""
 from .backend import use, active_backend, available_backends
-from .lang import index, indices, tensor, template, _0, _1, delta
+from .lang import index, indices, tensor, template, _0, _1, delta, pi
 from .lang._math import *  # noqa: F401, F403
 from .model import Factorization
 from .algorithm import factorize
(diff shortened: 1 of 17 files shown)


campsd:
Looks good to me. 

Two remarks:

- Regarding the missing slice- and element-wise access: Didn't we already have this for Kronecker product indices and can we use that here as well?
- Should we use the same notation for Kronecker product and Kronecker indices? `a & b` and `a[[&i, j]] * b[[&i, j]]`? This may be more intuitive.

yhtang:
> * Regarding the missing slice- and element-wise access: Didn't we already have this for Kronecker product indices and can we use that here as well?

I think so, it's just that we need to relocate the code and tweak them so that they work under the new primitive.

> * Should we use the same notation for Kronecker product and Kronecker indices? `a & b` and `a[[&i, j]] * b[[&i, j]]`? This may be more intuitive.

Unfortunately, `&` is not a unary operator in Python. Our life would be easier if this was C++ :joy:

yhtang:
I think we can tentatively go with this syntax and see how it feels in the real world.

PR #127, opened 2022-01-04
Jupyter notebook demo of SVD/NMF/RBF approximation

yhtang:
Add an example notebook which is directly rendered as an 'example' page in the online documentation.


yhtang:
@campsd 
I'm gonna merge this one so that the mkdocs setup is available for your QC notebook.

Issue #84, opened 2021-12-19
[BUG] `dtype`  not used in tensor initializer.


yhtang:
**Describe the bug**
The `LeafInitializer` interpreter has an optional argument `dtype` which defaults to `ab.float32`. However, it is not used by any of the primitives.

**Expected behavior**
The data type of the created tensor should be specified by this argument.

**Additional context**
Having `ab.float32` as the default value in the function signature causes very early import of the default backend. This is an undesired behavior. Since the initializer is very likely an internal tool used only by developers, I suggest we remove the default values.

yhtang:
@campsd pointed out that this will be highly relevant to the quantum compilation application.

campsd:
Collecting some thoughts and findings:

* [Gotcha #5](https://github.com/google/jax#current-gotchas) from JAX is something to take into account here. By default JAX only supports single precision
* [Complex single precision](https://jax.readthedocs.io/en/latest/notebooks/autodiff_cookbook.html#complex-numbers-and-differentiation) appear to be fully supported
* [PyTorch](https://pytorch.org/docs/stable/complex_numbers.html) has full support

yhtang:
> Gotcha #5 from JAX is something to take into account here. By default JAX only supports single precision

This seems quite manageable to me. We can insert the `JAX_ENABLE_X64=True` environment variable if, say, user do `ff.use('jax64')`.

campsd:
I ran some tests and that seems to work. However, the environment variable has to set before JAX is loaded.

yhtang:
> However, the environment variable has to set before JAX is loaded.

Can we add an `enable_complex=True/False`option to `backend.use` that auyomatically does this for the users?

campsd:
> > However, the environment variable has to set before JAX is loaded.
> 
> Can we add an `enable_complex=True/False`option to `backend.use` that auyomatically does this for the users?

Yes, that could work. However it would have to be `enable_64bit=True/False`.

yhtang:
> > > However, the environment variable has to set before JAX is loaded.
> > 
> > 
> > Can we add an `enable_complex=True/False`option to `backend.use` that auyomatically does this for the users?
> 
> Yes, that could work. However it would have to be `enable_64bit=True/False`.

Ah I see. My mind was still going with the complex number stuff. :joy: 

BTW this option mentioned on JAX docs seems better than the environment variable option:
```
# again, this only works on startup!
from jax.config import config
config.update("jax_enable_x64", True)
```
[...]

PR #128, opened 2022-01-05
Support for dtype

campsd:
Bugfix for #84.

Example code for complex optimization:

```python
import funfact as ff
import numpy as np
ff.use('torch')
m = np.random.rand(4,2,2)
m = m[..., 0] + 1j * m[..., 1]
n = np.random.rand(2,3,2)
n = n[..., 0] + 1j * n[..., 1]
target = m @ n
a = ff.tensor(4,2)
b = ff.tensor(2,3)
i,j,k = ff.indices(3)
tsrex = a[i,j] * b[j,k]
fac = ff.factorize(tsrex, target, nvec=64, max_steps=10000, dtype='target')
```

### DIFF
--- a/funfact/algorithm.py
+++ b/funfact/algorithm.py
@@ -1,7 +1,6 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 import numpy as np
-import operator
 import tqdm
 import funfact.optim
 import funfact.loss
@@ -74,10 +73,10 @@ def factorize(
     '''
 
     if dtype is None:
-        dtype = str(target.dtype)
-        dtype = operator.attrgetter(dtype)(ab)
-
-    target = ab.tensor(target, dtype=dtype)
+        target = ab.tensor(target)
+        dtype = target.dtype
+    else:
+        target = ab.tensor(target, dtype=dtype)
 
     @ab.autograd_decorator
     class _Factorization(Factorization, ab.AutoGradMixin):
--- a/funfact/algorithm.py
+++ b/funfact/algorithm.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 import numpy as np
+import operator
 import tqdm
 import funfact.optim
 import funfact.loss
@@ -12,7 +13,7 @@ from funfact.vectorization import vectorize, view
 def factorize(
     tsrex, target, lr=0.1, tol=1e-6, max_steps=10000, optimizer='Adam',
     loss='mse_loss', nvec=1, stop_by='first', returns='best',
-    checkpoint_freq=50, **kwargs
+    checkpoint_freq=50, dtype='target', **kwargs
 ):
     '''Factorize a target tensor using the given tensor expression. The
     solution is found by minimizing the loss function between the original and
@@ -55,6 +56,13 @@ def factorize(
             - If 'all', return all instances.
 
         checkpoint_freq (int >= 1): The frequency of convergence checking.
+
+        dtype: The datatype of the factorization model ('target', ab.dtype):
+
+            - If 'target', the same data type as the target tensor is used.
+            - If concrete dtype (float32, float64, complex64, complex128),
+            that data type is used.
+
     Returns:
         *:
             - If `returns == 'best'`, return a factorization object of type
@@ -65,7 +73,11 @@ def factorize(
             that represents all the solutions.
     '''
 
-    target = ab.tensor(target)
+    if dtype == 'target':
+        dtype = str(target.dtype)
+        dtype = operator.attrgetter(dtype)(ab)
+
+    target = ab.tensor(target, dtype=dtype)
 
     @ab.autograd_decorator
     class _Factorization(Factorization, ab.AutoGradMixin):
@@ -96,7 +108,7 @@ def factorize(
             )
 
     tsrex_vec = vectorize(tsrex, nvec)
-    opt_fac = _Factorization.from_tsrex(tsrex_vec)
+    opt_fac = _Factorization.from_tsrex(tsrex_vec, dtype=dtype)
 
     try:
         opt = optimizer(opt_fac.factors, lr=lr, **kwargs)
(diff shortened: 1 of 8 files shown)


funfact/algorithm.py
```diff
@@ -65,7 +73,11 @@ def factorize(
             that represents all the solutions.
     '''
 
-    target = ab.tensor(target)
+    if dtype == 'target':
+        dtype = str(target.dtype)
+        dtype = operator.attrgetter(dtype)(ab)
```
yhtang, line 78: I suppose this is, for example, to convert numpy dtypes to the equivalent ones in the backend?

How about flipping the order of this `if` block with line 80, so that we can directly use the type of the new `target` tensor?

funfact/backend/__init__.py
```diff
@@ -55,6 +59,8 @@ def use(backend: str):
     '''
     global _active_backend
     try:
+        if enable_x64 and backend == 'jax':
+            os.environ['JAX_ENABLE_X64'] = 'True'
```
yhtang, line 63: Can we also verify if x64 is enabled using the method as suggested by the JAX [docs](https://jax.readthedocs.io/en/latest/notebooks/Common_Gotchas_in_JAX.html):
> We can then confirm that x64 mode is enabled:
``` py
import jax.numpy as jnp
from jax import random
x = random.uniform(random.PRNGKey(0), (1000,), dtype=jnp.float64)
x.dtype # --> dtype('float64')
```

funfact/backend/_torch.py
```diff
@@ -20,11 +20,11 @@ def tensor(cls, array, optimizable=False, **kwargs):
         return torch.tensor(array, requires_grad=optimizable, **kwargs)
 
     @classmethod
-    def to_numpy(cls, tensor):
+    def to_numpy(cls, tensor, **kwargs):
         if tensor.requires_grad:
-            return tensor.detach().numpy()
+            return np.asarray(tensor.detach().numpy(), **kwargs)
```
yhtang, line 25: This can be made shorter by
```
if tensor.requires_grad:
    tensor = tensor.detach()
return np.asarray(...)
```

funfact/loss.py
```diff
@@ -45,7 +45,7 @@ def __call__(self, model, target, reduction='mean', sum_vec=True,
 class MSE(Loss):
 
     def _loss(self, model, target):
-        return ab.square(ab.subtract(model, target))
+        return ab.square(ab.abs(ab.subtract(model, target)))
```
yhtang, line 48: Maybe leave a comment to say that `abs` is for handling complex data? Otherwise, it may appear as redundant given the `square` operation.

funfact/algorithm.py
```diff
@@ -12,7 +13,7 @@
 def factorize(
     tsrex, target, lr=0.1, tol=1e-6, max_steps=10000, optimizer='Adam',
     loss='mse_loss', nvec=1, stop_by='first', returns='best',
-    checkpoint_freq=50, **kwargs
+    checkpoint_freq=50, dtype='target', **kwargs
```
yhtang, line 16: How about `dtype=None`, and then infer `dtype` if it is None? This should be as intuitive as `'target'`, and since infer dtype from target is the only special case here, we don't really need to use a string to name it.

funfact/algorithm.py
```diff
@@ -55,6 +56,13 @@ def factorize(
             - If 'all', return all instances.
 
         checkpoint_freq (int >= 1): The frequency of convergence checking.
+
+        dtype: The datatype of the factorization model ('target', ab.dtype):
+
+            - If 'target', the same data type as the target tensor is used.
```
yhtang, line 62: If `None`

yhtang (changes requested):

funfact/backend/__init__.py
campsd: I have added this to `_jax.py`. However, this warning will only trigger if JAX was already loaded separately from FunFact before with a different X64 setting. E.g.
```
import jax # load 32 bit
...
ff.use('jax', enable_x64=True) # warning will be shown here
```

 If the user does something like:
```
ff.use('jax', enable_x64=False) # load 32 bit
...
ff.use('jax', enable_x64=True) # try to load 64 bit
```
The second run does not actually get imported again presumably because in
```
_active_backend = getattr(
            importlib.import_module(f'funfact.backend._{backend}'), clsname
        )
```
`_jax.py` has already been loaded before. So the warning is not thrown in this case.

funfact/backend/_torch.py
campsd: Done.

funfact/loss.py
campsd: Done.

funfact/algorithm.py
campsd: Done

funfact/algorithm.py
campsd: The idea by first converting it to a native type tensor was that `dtype = str(target.dtype)` always gives a dtype from the backend. 

For example with torch this will become something like `'torch.float32'` which can be found by `operator.attrgetter`. If we flip it around, we can get something like `'numpy.float3'` which would cause an error in line 78 as it is not an attr of `ab`.

funfact/algorithm.py
campsd: Done

funfact/algorithm.py
yhtang: Oh, I meant what if we do this:
``` py
if dtype is None:
    target = ab.tensor(target)
    dtype = target.dtype
else:
    target = ab.tensor(target, dtype=dtype)
    dtype = target.dtype
```
and then be done. If the user wanna explicit control of dtype, they will manually specify a dtype from `ab`, or convert target to `ab.native_t` before passing it into vectorize.

funfact/backend/__init__.py
yhtang: I see. Sounds really tricky. How about we just stay as-is and assume that users will do things in the right order for the time being then? :cry:

yhtang (approved):

yhtang:
Could you also please check on the failed tests? :smile:

funfact/backend/__init__.py
campsd: Ok, I have removed this again.

campsd:
The tests are working now.

I'll look into that JAX complex conjugate issue in a separate PR.

yhtang (approved):

yhtang:
> 

Cool. Thanks!

PR #130, opened 2022-01-06
conjugate(grad) for JAX

campsd:
JAX returns the complex conjugate of the gradient by default. This PR fixes the JAX backend to return the true complex gradient.


campsd:
This is the intended behavior for JAX, see [here](https://github.com/google/jax/issues/9110) and [here](https://github.com/google/jax/issues/4891).

PR #131, opened 2022-01-06
Post and pre order vectorization, vectorized initializers

campsd:


campsd:
I'm not sure if we would want to expose the `post` order flag to the users in factorize. @yhtang  what do you think?

funfact/initializers.py
```diff
@@ -125,3 +125,22 @@ def __call__(self, shape):
         shape = as_tuple(shape)
         std = (self.scale / shape[self.axis])**0.5
         return std * self.distribution(shape)
+
+
+def vectorize_initializer(initializer, post: bool = True):
```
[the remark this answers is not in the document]
yhtang: Also, 'vectorize_initializer' might be a bit too wordy to type. I understand that this is to avoid conflict with the `vectorize` method for the tensor expressions. How about we steal `vmap` from JAX? Or maybe call it `stack` since we are literally stacking up lower dimensionality outputs?

funfact/lang/_terminal.py
```diff
@@ -165,13 +166,14 @@ def __init__(self, *size, symbol=None, initializer=None, optimizable=True):
         self.initializer = initializer
         self.optimizable = optimizable
 
-    def vectorize(self, n):
+    def vectorize(self, n, post):
```
yhtang, line 169: As above, maybe change `post` to `axis`?

funfact/initializers.py
```diff
@@ -125,3 +125,22 @@ def __call__(self, shape):
...
+    '''Vectorizes an initializer.
+
+    Args:
+        post (bool):
+            If True, the last index of shape is considered the vectorizing
+            index. If False, the first index of shape tuple is considered
+            the vectorizing index.
+    '''
+    if initializer is None:
```
yhtang, line 139: I understand that this is to handle the case when an abstract tensor has `None` as the initializer.

However, I think it might be semantically more straightforward if this is handled at the terminal level (AbstractTensor), since vectorizing None is not a very meaningful action.

What we can do is to simply do nothing in `AbstractTensor.vectorize` if the initializer is None, because in that case it'll default to `Normal`, which does not need vectorization.

funfact/initializers.py
```diff
@@ -125,3 +125,22 @@ def __call__(self, shape):
...
+        return None
+
+    def wrapper(shape):
+        return ab.stack([initializer(shape[:-1] if post else shape[1:]) for i
```
yhtang, line 143: It's better to factor our some calculations before we invoke list comprehension, as if it is too long then we lose the readability benefit.

funfact/lang/interpreter/_vectorize.py
```diff
@@ -47,17 +47,20 @@ class LeafVectorizer(_VectorizerBase):
 
     as_payload = TranscribeInterpreter.as_payload
 
-    def __init__(self, replicas: int, vec_index: P.index):
+    def __init__(self, replicas: int, vec_index: P.index, post: bool = True):
```
yhtang, line 50: Maybe `post` -> `axis`?

funfact/lang/interpreter/_vectorize.py
```diff
@@ -84,8 +88,14 @@ def ein(
         self, lhs: P.Numeric, rhs: P.Numeric, precedence: int, reduction: str,
         pairwise: str, outidx: Optional[P.indices], live_indices, **kwargs
     ):
+        if self.post:
```
yhtang, line 91: Since the branching is strictly binary here, I think it is better to use `else` for the other return statement.

yhtang: And perhaps factor out the common `*[P.index...]` part and only handles the appending/prepending part in the branches?

funfact/vectorization.py
```diff
@@ -9,7 +9,7 @@
 from funfact.model import Factorization
 
 
-def vectorize(tsrex, n):
+def vectorize(tsrex, n, post: bool = True):
```
yhtang, line 12: `post` -> `axis`?

yhtang (changes requested):

yhtang:
> I'm not sure if we would want to expose the `post` order flag to the users in factorize. @yhtang what do you think?
> 
> _Edit:_ After some more thought I think that it probably makes sense to change the default to false and let the user change this if required. Having the vectorizing index first easily allows batched application of numerous NumPy functions.

I agree. Since the convention is to prepend the vectorizing dimension, let's just conform to that.

p.s. Maybe this is a sign that I'm no longer as young and rebellious... :grin:

funfact/initializers.py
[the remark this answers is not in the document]
campsd: What do you think about renaming it to 'append' but keeping it as boolean? Since we only ever have two options I think a boolean makes most sense, or do you think there will ever be a reason to set an intermediate axis?

funfact/lang/_terminal.py
campsd: See reply above, we can change it, but I feel like a boolean is appropriate?

funfact/initializers.py
campsd: Done.

funfact/lang/interpreter/_vectorize.py
campsd: Done.

funfact/initializers.py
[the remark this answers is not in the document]
yhtang: Yeah this sounds like a good choice!

yhtang (approved):
Nice!

Issue #119, opened 2021-12-30
Support Ellipsis in indices?


yhtang:
**Is your feature request related to a problem? Please describe.**
Right now, users have to explicitly provide all the indices to perform Einstein operations. In cases where simple broadcasting is obvious, it might be more convenient to allow ellipsis indexing. `numpy.einsum` already supports this in the einspec strings.

**Describe the solution you'd like**
```
a = ff.tensor('a', 1, 2, 3, 4)
b = ff.tensor('b', 3, 2, 4, 6)
# current
i, j, k, r, s, p, q = ff.indices('i, j, k, r, s, p, q')
a[i, j, k, r] * b[p, q, r, s]
# proposed
i, j, k = ff.indices('i, j, k')
a[..., i, j] * b[..., j, k]
```

yhtang:
A possible approach to implement it would be to introduce a special value for the `indices` AST primitive, and then rewrite it using lhs&rhs information in the `__call__` of the new index propagator.

yhtang:
Per discussion with @campsd, it might be better to collect more usage scenarios before deciding on to what extent should this feature be supported.

PR #132, opened 2022-01-07
[WIP] Ellipsis support in index notations

campsd:
WIP towards #119 

This approach just leaves an Ellipsis as part of the index notations.
It renders a tsrex with an Ellipsis as follows:

```python
a = ff.tensor('a', 3, 2, 3, 4)
b = ff.tensor('b', 3, 2, 4, 6)
tsrex = a[..., i, j] * b[..., j, k]
tsrex
```

$a_{{\ldots}{i}{j} \times b_{{\ldots}{j}{k}}$

I think a viable approach to evaluate this is to change the `EinsteinSpecGenerator` to create the following string for the above tensor expression: `'...ab,...bc->...ac|'` and add support for these types of spec strings to `einop`.

The shape and syntax analyzers would also need to be updated.

What do you think of this approach @yhtang ?

### DIFF
--- a/funfact/lang/_tsrex.py
+++ b/funfact/lang/_tsrex.py
@@ -17,7 +17,12 @@ from .interpreter import (
     ShapeAnalyzer,
     EinsteinSpecGenerator
 )
-from ._terminal import LiteralValue, AbstractIndex, AbstractTensor
+from ._terminal import (
+    LiteralValue,
+    AbstractIndex,
+    AbstractEllipsis,
+    AbstractTensor
+)
 
 
 class ASCIITreeFactory:
@@ -345,16 +350,10 @@ def _getitem(node: _ASNode, indices):  # noqa: F811
     #                         'Indices to a tensor expression must be '
     #                         'abstract indices.'
     #                     )
-    new_indices = tuple([i.root if i is not Ellipsis else Ellipsis for i in
-                         as_tuple(indices or [])])
-    return P.index_notation(
-        node,
-        P.indices(new_indices)
-    )
-    # return P.index_notation(
-    #     node,
-    #     P.indices(tuple([i.root for i in as_tuple(indices or [])]))
-    # )
+    indices = tuple([i.root if i is not Ellipsis else
+                     P.ellipsis()
+                     for i in as_tuple(indices or [])])
+    return P.index_notation(node, P.indices(indices))
 
 
 @_dispatch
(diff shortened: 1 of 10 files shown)


yhtang:
Maybe an easier approach is to have a separate interpretation pass that **translates** ellipses into anonymous indices. That way, the rest of our pipeline can remain unchanged. The cons is that the translation will compromise readability --- anonymous indices like `#_1` are lengthy and not particularly aesthetic.

That said, I am not sure which approach will end up using fewer LOCs. If all that we need to change are concentrated in `einop`, then maybe adding specialization for ellipsis as you suggested could be quite practical.

funfact/lang/_tsrex.py
```diff
@@ -345,10 +345,16 @@ def _getitem(node: _ASNode, indices):  # noqa: F811
     #                         'Indices to a tensor expression must be '
     #                         'abstract indices.'
     #                     )
+    new_indices = tuple([i.root if i is not Ellipsis else Ellipsis for i in
```
yhtang, line 348: We can overwrite (rebind) `indices` here since the original one is no longer used after this line, and also since this is a super-small scope.

That said, a bigger concern here is the inclusion of `Ellipse`, which is not of `_ASNode` type, in the list of indices. This could potentially complicate future code development since we will then always need to check if a child is a node or not.

We can probably either:
1. add an ellipsis primitive, or
2. create a special index node.

Option 2 in my opinion may be deceptively simple but could cause problems once users start to decorate the special index with `~` or `*`, etc. :grin: So perhaps option 1 is really the way to go in this case.

yhtang: After a second thought, I am still not sure if a literal Ellipsis or an ellipsis primitive is better. Do you have any further thoughts?

campsd:
Going to postpone and abandon this for now.


Few notes when we take this up again later:

- We can probably fix the einspec and shape interpreters probably by inserting temporary local indices for the Ellipsis.
- adding the ellipsis should have the same effect as specifying all indices with a tilde.
- ellipsis can also be used with the `>>` operator to reorder output indices.

funfact/lang/_tsrex.py
campsd: I'm looking into this feature again. 

If we go with either option 1 or 2, we have to link the `...` (`Ellipsis`) notation directly with an `_ASNode` type object (be it a primitive or special index node) *before* `_getitem` is called while creating a tensor expression.

Any thoughts on how to achieve this?

campsd: A solution that I see is to create a function `_as_index` that is called on the indices in the `_getitem` method and replaces an `Ellipsis` with this special index node (or primitive object). 

Would this address the concern you raised above?

> That said, a bigger concern here is the inclusion of Ellipse, which is not of _ASNode type, in the list of indices. This could potentially complicate future code development since we will then always need to check if a child is a node or not.

PR #133, opened 2022-01-07
Special tensors. Closes #123.

### LINKED ISSUE #123: [REQ] Identity matrix
**Is your feature request related to a problem? Please describe.**
While the identity matrix can be regarded as a special realization of the Kronecker delta tensor, it may still be worthy of a direct API due to its wide usage in quantum computing, etc.

**Describe the solution you'd like**
``` py
# creates a 3x3 identity matrix with the symbol `I`
# `I` should have a special initializer that creates a sparse matrix instead of a dense one.
I = ff.eye(3) 
```

yhtang:
Added the `zeros`, `ones` and `eye` special tensors, and also the `Eye` initializer.

``` py
import funfact as ff
ff.ones(3, 4)
ff.zeros(4)
ff.eye(3)
```


funfact/initializers.py
```diff
@@ -30,6 +30,23 @@ def __call__(self, shape):
...
+    '''Initializes diagonal elements to 1 and all others to 0.
+
+    Args:
+        dtype: Numerical type of elements.
+    '''
+    def __init__(self, dtype=None):
+        self.dtype = dtype or ab.float32
+
+    def __call__(self, shape):
+        if len(shape) != 2 or shape[0] != shape[1]:
```
campsd, line 43: the `eye` methods in torch and numpy both support rectangular identity matrices with a second input argument (m and M respectively), We could also support that by forwarding this to the backends?

funfact/lang/_special.py
```diff
@@ -0,0 +1,53 @@
...
+        )
+    )
+
+
+def delta(n, d, optimizable=False, dtype=None):
+    return TsrEx(
+        P.tensor(
+            AbstractTensor(
+                n, n, symbol=(r'\delta', None),
+                initializer=ini.KroneckerDelta(dtype=dtype),
```
campsd, line 49: I don't see this initializer yet? Maybe leave a TODO note that we still need to add it?

funfact/initializers.py
yhtang: Sure!

funfact/lang/_special.py
yhtang: I later decided to remove the initializer since we don't have a concrete use case yet, but forget to remove the special abstract tensor. It is now been removed.

PR #141, opened 2022-01-10
full support for prepend vectorization

campsd:
This adds complete support for prepending vectorization to:

- FF loss functions
- vectorization.view 
- factorize
- backends

The following code now works correctly:

```python
A = ff.tensor('A', 3, 2)
B = ff.tensor('B', 2, 3)
#tsrex = A[[*i, *j]] * B[[i, j]]
tsrex = A[[i, j]] * B[[j, k]]
target = ff.Factorization.from_tsrex(tsrex)
target = target()
fac = ff.factorize(tsrex, target, nvec=4, append=False)
fac()
```

### DIFF
--- a/funfact/algorithm.py
+++ b/funfact/algorithm.py
@@ -92,10 +92,10 @@ def factorize(
                 'funfact.loss.'
             )
     try:
-        loss(target, target, **kwargs)
+        loss(target, target, append=append, **kwargs)
     except Exception as e:
         raise AssertionError(
-            f'The given loss function does not accept two arguments:\n{e}'
+            f'The given loss function does not accept three arguments:\n{e}'
         )
 
     if isinstance(optimizer, str):
@@ -117,7 +117,7 @@ def factorize(
             'Invalid optimization algorithm:\n{e}'
         )
 
-    loss_and_grad = ab.loss_and_grad(loss, opt_fac, target)
+    loss_and_grad = ab.loss_and_grad(loss, opt_fac, target, append=append)
 
     # bookkeeping
     best_factors = [np.zeros_like(ab.to_numpy(x)) for x in opt_fac.factors]
@@ -131,11 +131,15 @@ def factorize(
 
             if step % checkpoint_freq == 0:
                 # update best factorization
-                curr_loss = ab.to_numpy(loss(opt_fac(), target, sum_vec=False))
+                curr_loss = ab.to_numpy(loss(opt_fac(), target, sum_vec=False,
+                                             append=append))
                 better = np.flatnonzero(curr_loss < best_loss)
                 best_loss = np.minimum(best_loss, curr_loss)
                 for b, o in zip(best_factors, opt_fac.factors):
-                    b[..., better] = o[..., better]
+                    if append:
+                        b[..., better] = o[..., better]
+                    else:
+                        b[better, ...] = o[better, ...]
 
                 converged |= np.where(curr_loss < tol, True, False)
                 if stop_by == 'first':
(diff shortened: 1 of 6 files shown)


funfact/algorithm.py
```diff
@@ -92,10 +92,10 @@ class _Factorization(Factorization, ab.AutoGradMixin):
                 'funfact.loss.'
             )
     try:
-        loss(target, target, **kwargs)
+        loss(target, target, append=append, **kwargs)
     except Exception as e:
         raise AssertionError(
-            f'The given loss function does not accept two arguments:\n{e}'
+            f'The given loss function does not accept three arguments:\n{e}'
```
yhtang, line 98: The 'two' arguments here refer to the target and approximate tensors that the loss function is meant to compare. Any additional arguments are 'meta/hyper-parameters' and hence does not count towards the number of arguments that a loss function accepts. 😁

funfact/loss.py
```diff
@@ -12,14 +12,15 @@ def _loss(self, model, target):
...
     def __call__(self, model, target, reduction='mean', sum_vec=True,
-                 **kwargs):
-        if target.ndim == model.ndim - 1:
-            if model.shape[:-1] != target.shape:
-                raise ValueError(f'Target shape {target.shape} and '
-                                 f'model shape {model.shape[:-1]} mismatch.')
-            data_axis = tuple(i for i in range(target.ndim))
-            target = target[..., None]
-        elif target.ndim == model.ndim:
+                 append=True, **kwargs):
```
yhtang, line 15: I find the meaning of `append` a bit cryptic here. This is different from the situation of the `vectorize` method, whose name already provides the context. Here, the context is 'loss', and hence it is not very intuitive as to what 'append' would mean. What about `vectorizes_along_last` or something along the aline?

funfact/vectorization.py
```diff
@@ -51,17 +51,21 @@ def view(fac, tsrex_scalar, instance: int):
             The original, un-vectorized tensor expression.
         instance (int):
             Index along the vectorization dimension.
+        append (bool):
```
yhtang, line 54: While I understand it is better to reuse the `append` keyword to ensure API consistency here, there is a similar semantic ambiguity problem as above in the loss function.

At least, we can change this docstring line to say something like 'indicates whether the vectorization dimension was appended or prepended'.

yhtang (changes requested):

yhtang:
Another issue here is what default values shall we assign to `append`. I recall that we want to be default-compatible with NumPy. In that case, perhaps we should make all defaults to `append=False`?

funfact/algorithm.py
campsd: Changed back.

funfact/vectorization.py
campsd: To have a consistent API, we can also call the argument 'vectorize_along_last' everywhere? Slight difference in meaning when it is an active action ('vectorize`) or details about input (`loss`) but maybe that's fine?

funfact/loss.py
campsd: Changed here to `vectorized_along_last`.

funfact/vectorization.py
campsd: I have updated the doc string, but kept the name.

campsd:
> Another issue here is what default values shall we assign to `append`. I recall that we want to be default-compatible with NumPy. In that case, perhaps we should make all defaults to `append=False`?

Good point, all the defaults should now be False everywhere.

yhtang (approved):

PR #145, opened 2022-01-11
Conditions and preferences for penalties on loss function

### LINKED ISSUE #143: Support penalty terms on leaf tensors
It would be good to have a mechanism to enforce penalties on the leaf tensors that are summed together with the loss when optimizing a tensor expression. A great example use case is a penalty on the unitariness of the leaf matrices in a quantum circuit, another similar example is if we want to optimize a matrix factorization with constraints on the leafs (e.g SVD, NNMF, QR)

It seems to me that the most straightforward way to add this feature is to let the user add a penalty term upon initialization:

An API somewhat like this:
```python
Q = ff.tensor(4, 4, initiazer=unitary_init, penalty=is_unitary)
R = ff.tensor(4, 2, penalty=is_upper_triangular)
tsrex = Q @ R
# ...
fac = ff.factorize(tsrex, target, penalty_weights=0.1) # optimize QR factorization of 4x2 matrix, penalty_weights can be the hyper parameter for the loss function.
```

I think this is more user-friendly and easier to add to FunFact (can we just use the payload mechanism?) compared to setting the penalties when calling the algorithm:

`factorize(tsrex, target, penalties={is_unitary: Q, is_upper_triangular: R})`

Because after some interpreter passes it might be harder to match the tensors in this case?

campsd:
~~WIP towards~~ Closes #143 

Implements some basics towards the penalty feature. <del>The current functionality:
```python
a = ff.tensor(2,3,penalty=ff.penalties.is_upper_triangular)
b = ff.tensor(3,4)
tsrex = a @ b
fac.penalty # will return a list with two values, second is 0 because no penalty is enforced on b
```
</del>

**EDIT:** The new syntax is:
```python
a = ff.tensor(2,3,prefer=ff.conditions.Unitary(weight=0.1, elementwise='l1', reduction='sum'))
b = ff.tensor(3,4)
tsrex = a @ b
fac.penalty
```

Some thoughts after working on this:

- I didn't add a `penalty` property to `BaseEx` as this would require running the initializer in the TsrEx class which we would like to avoid because of lazy evaluation
- Instead added this to the `Factorization` model. Not sure if a list (or `NodeView`?) is the best output or if we want to reduce/sum all penalties here?

~~Still working on:~~
- ~~vectorizing the penalties~~
- ~~more penalty functions~~
- ~~taking the penalty into account in the `factorize` algorithm~~

**EDIT:** It is now fully compatible with both backends and the penalties are taken into account during the optimization process.


yhtang:
> * I didn't add a `penalty` property to `BaseEx` as this would require running the initializer in the TsrEx class which we would like to avoid because of lazy evaluation

Sounds very reasonable --- a TsrEx is an 'abstract' expression and hence we probably shouldn't attempt to compute concrete penalty values on it anyway.

> * Instead added this to the `Factorization` model. Not sure if a list (or `NodeView`?) is the best output or if we want to reduce/sum all penalties here?

IMHO a list is perfect, and the users can perform a summation very trivially if that's what's needed. Node views are not necessary though, since it was designed to enable writable access to tensor data. The penalties are read-only states of the tensors.

yhtang:
Our previous discussion was that 'constraints', in particular **hard** constraints, are usually meant to be satisfied rigorously, and since SGD can likely only deliver a relaxed approximation to the conditions, we decided to use the word 'penalty' term to emphasize the 'soft' nature of the conditions.

However, I find it a bit awkward to say 'the penalty for matrix A is upper triangularity' because it sounds as if the matrix is penalized for being upper triangular, which is exactly the opposite of what actually happens.

How about we use the word `prefer`? e.g.
```
a = ff.tensor('a', 2, 3, 4, prefer=ff.conditions.upper_triangular)
b = ff.tensor('b', 5, 5, prefer=ff.conditions.diagonal)
```
This should simultaneously emphasize the soft nature of the conditions while also ensuring straightforward semantics.

Let me know your thoughts :grin:

funfact/lang/_terminal.py
```diff
@@ -165,9 +166,11 @@ def __init__(self, *size, symbol=None, initializer=None, optimizable=True):
         self.symbol = self.TensorSymbol(symbol or self.uuid)
         self.initializer = initializer
         self.optimizable = optimizable
+        self._penalty = penalty
```
yhtang, line 169: I think we can leave it as a public member, just like `initializer` and `optimizable`.

funfact/model/_factorization.py
```diff
@@ -133,6 +134,14 @@ def ndim(self):
         '''The dimensionality of the result tensor.'''
         return self.tsrex.ndim
 
+    @property
+    def penalty(self):
+        '''The penalty of the result tensor.'''
+        _tsrex = self.tsrex | PenaltyEvaluator()
```
yhtang, line 140: No need for the `_` prefix here since these variables are local within the scope of the method. :smile:

campsd:
I like the terminology of `prefer=` and renaming the penalties module to `conditions`. Will make these adjustments. 

How should we rename the `PenaltyEvaluator` interpreter? `PreferenceChecker`, `PreferenceEvaluator`, `ConditionChecker`, ...? 😄

campsd:
It now works for the JAX backend. There's probably a lot of fine-tuning required to get it working better.
```python
A = np.random.randn(6,4)
U = ff.tensor('U', 6, 4, prefer=ff.conditions.Unitary(weight=0.5))
S = ff.tensor('S', 4, 4, prefer=ff.conditions.Diagonal())
V = ff.tensor('V', 4, 4, prefer=ff.conditions.Unitary())
tsrex = U[i,j] * (S[j, k] * V[k, l])
fac = ff.factorize(tsrex, A, nvec=10, tol=5e-3, penalty_weight=0.1)
fac.penalty(sum_leafs=False) # show how well the penalties are satisfied
```
For some yet unknown reason it doesn't work with PyTorch. If the penalty are not evaluated by setting `penalty_weight=0.0` it converges, but a positive weight doesn't converge. It seems like it is not tracing this correctly or something. Any idea what could be going wrong?

funfact/algorithm.py
```diff
@@ -117,7 +119,15 @@ class _Factorization(Factorization, ab.AutoGradMixin):
             'Invalid optimization algorithm:\n{e}'
         )
 
-    loss_and_grad = ab.loss_and_grad(loss, opt_fac, target,
+    def loss_and_penalty(model, target, sum_vec=True, **kwargs):
+        if penalty_weight > 0:
```
yhtang, line 123: Based on the DRY (don't repeat yourself) principle, it is probably better to first compute the loss, and then conditionally compute the penalty term. This way, we don't repeat the call to the loss function.

funfact/algorithm.py
```diff
@@ -132,8 +142,13 @@ class _Factorization(Factorization, ab.AutoGradMixin):
 
             if step % checkpoint_freq == 0:
                 # update best factorization
-                curr_loss = ab.to_numpy(loss(opt_fac(), target, sum_vec=False,
-                                             vectorized_along_last=append))
+                curr_loss = ab.to_numpy(
+                    loss_and_penalty(
+                        opt_fac, target, sum_vec=False,
+                        vectorized_along_last=append,
+
```
yhtang, line 149: Is this a stray empty line?

funfact/initializers.py
```diff
@@ -148,6 +148,7 @@ def stack(initializer, append: bool = True):
     '''Stacks initializers for the purpose of vectorization.
 
     Args:
+        initializer (callable): non-vectorized initializer.
```
yhtang, line 151: 'non-vectorized' sounds a bit awkward to me. How about 'The initializer to the vectorized'?

funfact/lang/_terminal.py
```diff
@@ -176,8 +179,10 @@ def vectorize(self, n, append):
         else:
             initializer = self.initializer[..., None] if append else \
                           self.initializer[None, ...]
+        prefer = vmap(self.prefer, append) if self.prefer else None
```
yhtang, line 182: Based on our discussion today, we can explore later whether it is possible to unify `stack` and `vmap`.

funfact/lang/_terminal.py
```diff
@@ -153,7 +154,8 @@ class TensorSymbol(Symbol):
         _anon_registry = {}
         _anon_registry_lock = multiprocessing.Lock()
 
-    def __init__(self, *size, symbol=None, initializer=None, optimizable=True):
+    def __init__(self, *size, symbol=None, initializer=None, optimizable=True,
+                 prefer=NoCondition()):
```
yhtang, line 158: I'd prefer `prefer` (prefer x2 :grin:) to have a default value of `None`, which gets translated into `NoCondition` later on . This helps to simplify the function signature and makes the documentation more readable, since users will not need to figure out what `NoCondition` means.

funfact/lang/_tsrex.py
```diff
@@ -415,7 +416,7 @@ def indices(spec):
         raise RuntimeError(f'Cannot create indices from {spec}.')
 
 
-def tensor(*spec, initializer=None, optimizable=None):
+def tensor(*spec, initializer=None, optimizable=None, prefer=NoCondition()):
```
yhtang, line 419: Similar here, `prefer=None` might be a more readable approach.

funfact/lang/interpreter/_penalty.py
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env python
```
yhtang, line 1: Since the penalty evaluator only operates on leaf nodes, we can simply use `dfs_filter` without defining a full-fledged interpreter. This could then be implemented as a standalone function.

funfact/model/_factorization.py
```diff
@@ -133,6 +135,21 @@ def ndim(self):
         '''The dimensionality of the result tensor.'''
         return self.tsrex.ndim
 
+    def penalty(self, sum_leafs: bool = True, sum_vec=None):
```
yhtang, line 138: Similar to the comment above, could we simply invoke the conditions on the `data` tensor while iterating over `dfs_filter`?

funfact/lang/_terminal.py
campsd: In the current implementation, there is actually a subtle difference between tensors that have `prefer=None` and the ones that have `prefer=NoCondition()`. The former is meant for non optimizable tensors for which the preference is not considered in the optimization, the latter is for optimizable tensors without a specified preference. Does that distinction make sense to you? I wanted to make this more explicit by changing line 22 of the `PenaltyEvaluator` to `return []`.

funfact/lang/interpreter/_penalty.py
campsd: Makes sense, although the interpeter isn't doing much more work compared to a `dfs_filter`. This was modeled after the leaf initializer. Probably the same comment applies to that interpreter as well?

yhtang: I see your point here. There is a key difference between the leaf initializer and the penalty evaluator: the former modifies (insert the `data` attribute to) the AST nodes, while the latter only performs read-only operations. That's why the leaf initializer could benefit from the TranscribeInterpreter infrastructure, while the penalty evaluator can be further simplified. :sunglasses:

funfact/lang/_terminal.py
yhtang: I see. Since constraints is not meaningful when a tensor is non-optimizable, could we check for the `optimizable` flag when evaluating the penalties, and simply return `tensor(0, dtype=...)` upon True? In the docstring we can just say that the `prefer` keyword is ignored if the optimizable flag is False. Similarly, if `prefer` is None, then we just return a 0 tensor without further actions.

funfact/algorithm.py
campsd: it is, removed.

funfact/lang/_terminal.py
campsd: ok, I have resolved this and made some changes in the `tensor` method in `_tsrex.py`. See comment there.

funfact/lang/_tsrex.py
campsd: Changed the default argument to `None`. Added
```   
 if optimizable and prefer is None:
        prefer = NoCondition()
```

as we need this for the `vmap` of optimizable leafs. The penalty should be a 1D array after vmap.

funfact/model/_factorization.py
campsd: Done.

funfact/lang/interpreter/_penalty.py
campsd: Makes sense, it is replaced now by a `dfs_filter` in the factorization model. In my preliminary tests this seems to improve the performance as well 😄

funfact/lang/_terminal.py
campsd: Makes sense, I'll create a separate issue for this.

PR #146, opened 2022-01-13
Redesign of index analyzer and vectorizer. Fixes #142.

### LINKED ISSUE #142: [BUG] Vectorization and index analyzer bug for more complicated tensor expressions
**Describe the bug**
The `vectorize` method throws `SyntaxError: Incorrect number of indices. Expects 4, got 3.` in the index analyzer pass for certain complicated tensor expressions.

**To Reproduce**
Steps or a code snippet to reproduce the behavior:
```python
from funfact.vectorization import vectorize
tsrex = ((ff.eye(2) & ff.eye(2)) + (ff.eye(2) & ff.eye(2)))
tsrex = tsrex & ff.eye(2)
tsrex_vec = vectorize(tsrex, 3)
```
The tensor expression on the second line can be correctly vectorized, adding the term on the 3rd line introduces the error.

yhtang:
Replaced the original index propagation and vectorization rewriting mechanism with a more rigorous one. In the new approach, tensor expressions are first created using 'abstract' primitives, which will be compiled into concrete ones upon evaluation. The compilation process consists of:
1. Inferring the 'indexness', i.e. interpreted as indexed or not, for each intermediate node. The rule of thumb is that a binary composite expression is indexed only if both of its operands are indexed. Otherwise, even if one of the operands may be indexed, the index will get lost after operations with another indexless tensor.
2. Merged the shape and live indices analysis into one interpreter, which also simultaneously rewrites a tensor expression by converting abstract primitives into concrete ones. The determination of the concrete type depends on the indexness, live indices status, and shape of the operands of a primitive. After rewriting, the newly created AST nodes may have to be re-evaluated for shape, live indices, etc.
3. Make sure a compiled tensor expression contains only elementwise and Einstein operations. Vectorizing can be then implemented by extending the dimensionality of the leaf tensors and indices by 1.


yhtang:
I am going to merge it in about an hour given the time frame that we are currently aiming at. However, please feel free to make comments after the merge, and we will address them in a new PR as long as necessary. :grin:

campsd:
There appears to be an issue with index transposition after the redesign. The following doesn't work:

```
a = ff.tensor('a', 4, 3, 2)
b = ff.tensor('b', 2, 4, 5)
tsrex = a[i, j, k] * b[k, i, l] >> [j, l]
```

PR #150, opened 2022-01-15
Fix bug when populating tensors from concrete initializers

yhtang:
@campsd discovered the problem:
``` py
import funfact as ff
import numpy as np
# instantiate data arrays:
a = np.array([1.0, 2.0, 3.0])
b = np.array([-1.0, 4.0])
# instantiate tensors from concrete data:
a = ff.tensor('a', a, optimizable=False)
b = ff.tensor('b', b, optimizable=True, prefer=ff.conditions.NonNegative())
i, j = ff.indices('i, j')
# create tensor expression and factorization model:
tsrex = a[i] * b[j]
```
which leads to `b` being a tensor of length 6. This is because of an incorrect usage of the `broadcast_to/tile` method.

### DIFF
--- a/funfact/lang/interpreter/_initialization.py
+++ b/funfact/lang/interpreter/_initialization.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
+import numpy as np
 from funfact.backend import active_backend as ab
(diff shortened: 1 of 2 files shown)


campsd (approved):
This looks good to me, it solves the bug 😄 thanks!

PR #153, opened 2022-01-17
Fix `initializers.stack` bug when classes are given as the initializer of tensors

yhtang:
Currently, `initializers.stack` assumes the underlying object is always a callable. This is inconsistent with our tsrex semantics, which allows initializer classes to be directly supplied (and we will instantiate the classes when necessary). This is now fixed by using a `StackedInitializer` class instead of the thinner wrapper function solution.

### DIFF
--- a/funfact/initializers.py
+++ b/funfact/initializers.py
@@ -154,9 +154,19 @@ def stack(initializer, append: bool = True):
             index. If False, the first index of shape tuple is considered
             the vectorizing index.
     '''
-    def wrapper(shape):
-        nvec = shape[-1] if append else shape[0]
-        shape = shape[:-1] if append else shape[1:]
-        axis = -1 if append else 0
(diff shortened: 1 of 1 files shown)


campsd:
Looks good. Alternatively this could maybe also be handled when the tensors are instantiated? If it the `initializer` is a class, it can already be initialized there. We could do the same for the `condition`.

yhtang:
> Looks good. Alternatively this could maybe also be handled when the tensors are instantiated? If the `initializer` is a class object, it can already be initialized there. We could do the same for the `condition`.

The tricky thing about that is the passing of `dtype` information, which must be passed through the `__init__` method of an initializer in our current design. This information, unfortunately, is not available at the time of `stack`.

I agree that we should consider this together with the conditions in a bigger context, given the similarities of the two modules.

PR #158, opened 2022-01-19
Unit tests for losses.

yhtang:
In addition to tests, I also relocated the `reduction` argument of the loss functions from `__call__` to `__init__`. `ff.factorize` was also updated to accept either class-valued or object-valued loss arguments.

### DIFF
--- a/docs/pages/user-guide/factorize.md
+++ b/docs/pages/user-guide/factorize.md
@@ -107,8 +107,7 @@ a custom optimizer:
 
 The loss function used in the optimization process can be changed as well as its
 (hyper) parameters. This adheres to the FunFact [Loss API](../../../api/loss/).
-All keyword arguments that are input to `factorize` are passed to the loss
-function.
+The `optimizer_options` dictionary argument to the `factorize` will be forwarded as keyword arguments to initialize the loss function.
 FunFact currently offers native support for three loss functions and allows
 for a custom loss function:
 
--- a/docs/pages/user-guide/factorize.md
+++ b/docs/pages/user-guide/factorize.md
@@ -105,20 +105,17 @@ a custom optimizer:
 
 ### Loss
 
-The loss function used in the optimization process can be changed as well as its
-(hyper) parameters. This adheres to the FunFact [Loss API](../../../api/loss/).
-The `optimizer_options` dictionary argument to the `factorize` will be forwarded as keyword arguments to initialize the loss function.
-FunFact currently offers native support for three loss functions and allows
-for a custom loss function:
+The loss function used in the optimization process can be changed as well as
+its (hyper) parameters. FunFact currently offers three loss functions:
 
-- `loss`: loss function:
-    * `'mse_loss'`: Mean-Squared Error (L2) loss.
-    * `'l1_loss'`: L1 loss.
-    * `'kldiv_loss'`: KL Divergence loss.
-    * `callable`: user provided optimizer that implements the [Loss 
-    API](../../../api/loss/).
+  * `'MSE'`: Mean-Squared Error (L2) loss.
+  * `'L1'`: L1 loss.
+  * `'KLDivergence'`: KL Divergence loss.
 
-The loss function can be further adjusted by the `penalty_weight` argument which
-is the scalar factor applied to all the penalty terms specified on the leaf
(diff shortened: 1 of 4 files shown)


docs/pages/user-guide/factorize.md
```diff
@@ -107,8 +107,7 @@ a custom optimizer:
 
 The loss function used in the optimization process can be changed as well as its
 (hyper) parameters. This adheres to the FunFact [Loss API](../../../api/loss/).
-All keyword arguments that are input to `factorize` are passed to the loss
-function.
+The `optimizer_options` dictionary argument to the `factorize` will be forwarded as keyword arguments to initialize the loss function.
```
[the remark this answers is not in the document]
yhtang: Right. I was probably drunk-coding then. It is now fixed (eliminated via the class-or-callable idiom). :grin:

funfact/algorithm.py
```diff
@@ -113,22 +115,23 @@ class _Factorization(Factorization, ab.AutoGradMixin):
     opt_fac = _Factorization.from_tsrex(tsrex_vec, dtype=dtype)
 
     try:
-        opt = optimizer(opt_fac.factors, lr=lr, **kwargs)
+        opt = optimizer(opt_fac.factors, lr=lr)
```
campsd, line 118: This means that we can not set the other hyperparameters of the optimizer? E.g. `beta1`, `beta2`, `epsilon` for Adam. Maybe we can use the dictionary approach to still allow this?

yhtang: I am thinking about adapting the class-or-object idiom to optimizers too. So if the user supplies a class (or str of the class name), it will be instantiated with the default arguments. Alternatively, they can create an optimizer object on their own so that additional arguments can be given there.

funfact/test_loss.py
```diff
@@ -0,0 +1,112 @@
...
+
+    loss = loss_cls()
+    for shape in [(3,), (3, 3), (3, 4), (3, 4, 5), (2, 3, 4, 5)]:
+        for _ in range(64):
+            a = ab.normal(0.0, 1.0, shape)
+            b = ab.normal(0.0, 1.0, shape)
+            assert loss(a, a) == pytest.approx(0, abs=1e-6)
+            assert loss(b, b) == pytest.approx(0, abs=1e-6)
+            assert loss(a, b) >= 0
+            assert loss(b, a) >= 0
```
campsd, line 26: Maybe we can add a test that `loss(a, b)` and `loss(b,a)` should be approximately equal?

funfact/test_loss.py
```diff
@@ -0,0 +1,112 @@
...
+    MSE,
+    L1,
+    KLDivergence
+])
+def test_reduction(loss_cls):
+
+    loss_sum = loss_cls(reduction='sum')
+    loss_mean = loss_cls(reduction='mean')
+    with pytest.raises(Exception):
+        loss_cls(reduction='abrakadabra')
```
campsd, line 109: Great reduction method 😆

funfact/algorithm.py
yhtang: In other words, we should avoid bringing the keywords of components such as loss or optimizer into the factorize method, because:
1. It will become more and more challenging if we involve more components. For example, I am eyeing multi-tensor penalty terms for V2.0 :grin: 
2. This would make the signature more and more complicated, and also requires us to duplicate docstrings.

campsd: Sounds like a good idea

funfact/test_loss.py
yhtang: Good point.

funfact/test_loss.py
yhtang: :sunglasses:

funfact/test_loss.py
campsd: Yes, but I believe this test only runs for Lp norms :)

yhtang: Done

campsd (approved):

funfact/test_loss.py
yhtang: > Yes, but I believe this test only runs for Lp norms :)

Yeah, I realized it as soon as I opened up the file. Seems like I already forgot part of what I wrote two days ago. :joy:

PR #162, opened 2022-01-21
Fix docs browser freeze due to example notebook.

yhtang:
Also tuned the RBF example notebook: reducing the learning rate to 0.05 for RBF leads to better results.


funfact/loss.py
```diff
@@ -99,3 +99,8 @@ class KLDivergence(Loss):
     '''KL Divergence loss.'''
     def _loss(self, model, target):
         return ab.multiply(target, ab.log(ab.divide(target, model)))
+
+
+mse = MSE()
```
campsd, line 104: What is the reason for reintroducing these?

yhtang: In case if the user (or me in this case) wants to directly use a loss function as a *function*, without the need to instantiate it first. :grin:

campsd (approved):

PR #163, opened 2022-01-21
Unit tests for `algorithm.py::factorize`

yhtang:
Along with some minor refactorizing of the code.

### DIFF
--- a/funfact/algorithm.py
+++ b/funfact/algorithm.py
@@ -77,11 +77,7 @@ def factorize(
 
     tsrex_vec = vectorize(tsrex, nvec, append=append)
 
-    @ab.autograd_decorator
-    class _Factorization(Factorization, ab.AutoGradMixin):
-        pass
-
-    opt_fac = _Factorization.from_tsrex(tsrex_vec, dtype=dtype)
+    fac = ab.add_autograd(Factorization).from_tsrex(tsrex_vec, dtype=dtype)
 
     if dtype is None:
         target = ab.tensor(target)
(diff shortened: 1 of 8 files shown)


funfact/algorithm.py
```diff
@@ -179,6 +179,6 @@ def loss_and_penalty(model, target, sum_vec=True):
             np.argsort(best_loss)[:returns]
         ]
     elif returns == 'all':
-        return best_fac
-    else:
-        raise RuntimeError(f'Invalid argument value for returns: {returns}')
+        return [
+            view(best_fac, tsrex, i, append) for i in np.argsort(best_loss)
```
campsd, line 183: I agree that it's good for consistent output that you changed this. Also update the doc string about `returns` being `'all'` and probably the user guide: https://funfact.readthedocs.io/en/latest/pages/user-guide/factorize/#vectorization-for-multi-instance-learning

funfact/test_algorithm.py
```diff
@@ -0,0 +1,110 @@
...
+    assert isinstance(fac, list)
+
+    fac = factorize(
+        tensor(2), ab.ones(2), nvec=4, max_steps=100, returns='all'
+    )
+    assert isinstance(fac, list)
+
+
+def test_penalty_weight():
+    fac = factorize(tensor(2), ab.ones(2), penalty_weight=1.0)
```
campsd, line 106: we could maybe also add a simple test with a penalty and check that it is satisfied?

campsd (changes requested):
Looks good, some minor updates to documentation required for `returns='all'`.

funfact/test_algorithm.py
yhtang: I was trying to refrain from doing it here, because as unit tests, they should involve as few components of the package as possible. Here we are just checking if the penalty is used at all by `factorize` per se. Whether the entire penalty mechanism works would then be left to the integration tests (which we have yet to build).

yhtang:
> Looks good, some minor updates to documentation required for `returns='all'`.

Done!

campsd (approved):

PR #165, opened 2022-01-22
Update the CFG docs page.

yhtang:


mkdocs.yml
```diff
@@ -79,7 +81,7 @@ nav:
       - Nonlinear matrix approximation: examples/matrix-approximation.ipynb
     - pages/cheatsheet.md
     - User Guide:
-        - pages/user-guide/tensor-and-indices.md
+        - "Tensors and Indices": pages/user-guide/tensor-and-indices.md
```
campsd, line 84: Do we need to add this type of descriptive string for the other pages of the user guide?

yhtang: This is the only place needed, to address a (IMHO) bug of `mkdocs` :joy: For other pages, the page's TOC title is extracted automatically from the H1 heading inside the document.

campsd (approved):

PR #169, opened 2022-01-24
optim unit tests

campsd:
unit tests for optim module


yhtang:
Could we implement a simple convergence test, e.g. by solving a quadratic problem with hand-derived gradient functions? :smile:

campsd:
Good idea, I'll add something along that line.

PR #172, opened 2022-01-25
Documentation for tensor and index creations. Closes #112.

### LINKED ISSUE #112: User guide: tensors and indices
`docs/pages/user-guide/tensor-and-indices.md`

This is the first part of the user guide. The goal is to:
- Introduce the APIs to create tensors and indices
- !!! note: explain the 'abstract' nature of the tensors, i.e. they are **not** yet populated with concrete numbers yet.
- !!! note: explain the index identity logic: identities check is performed on UUIDs on a Python variable basis, not done on the 'symbol', which is merely for human visualization.

yhtang:


docs/pages/user-guide/tensor-and-indices.md
```diff
@@ -2,24 +2,75 @@
...
+prevalence in linear algebra computations. These include:
+
+- The [unit tensor](../../../api/ones) of all 1s.
+- The [zero tensor](../../../api/zeros) of all 0s.
+- The [identity/Kronecker delta tensor](../../../api/eye). 
+
+## Initializers
+
+By default, the abstract tensors are populated with random elements draw
+from the normal distribution upon initialization of a factorization model. A
```
campsd, line 48: Should we mention that by default the stdev of the distribution is very small?

campsd (changes requested):
Looks good. I've left two small comments.

docs/pages/user-guide/tensor-and-indices.md
yhtang: I was thinking about a future PR that relocates the docstrings for the initializers from class level to `__init__`. This way, the default values will be correctly presented on the API page. And then we can have a 'sharp bits' page to consolidate this type of information.

docs/pages/user-guide/tensor-and-indices.md
```diff
@@ -2,24 +2,75 @@
...
+``` py title="Examples"
+import ff.initializers as ini
+X = ff.tensor('X', 5, 2, 4, initializer=ini.Uniform)
+Y = ff.tensor('Y', 5, 2, 4, initializer=ini.Uniform(scale=1.25))
+Z = ff.tensor('Z', 5, 2, 4, initializer=ini.Normal(mean=-1, std=0.2))
+```
+
+## Preferences and conditions
+
+Many factorization algorithms enforces constraints on the factors found in the solution.
```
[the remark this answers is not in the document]
yhtang: Done :grin:

PR #177, opened 2022-01-26
factorization tests

campsd:
100% coverage for factorization module.


yhtang (changes requested):

campsd:
Updated unit tests w/o loading the whole package, only relying on the `ab`
.

yhtang (approved):

Issue #179, opened 2022-01-28
[REQ] Repeated calculation of transposition-pairwise-reduction-reshape parameters in `einop` is unnecessary


yhtang:
Currently, the `_einop` method takes in an `einspec` string and uses expensive inference logic to compute how the operand tensors should be transposed in order to perform the contraction. However, in our scenario where a tsrex is reused repeatedly in an optimization loop, this calculation could have been done at compilation time, thus saving time.

Hence, we should eliminate the `EinspecGenerator` interpreter, which seems to take a detour from creating NumPy-style einspec strings, and create a new `EinopCompiler` interpreter that directly returns the _transposition_, _axes expansion_, and _reshaping_ parameters as the payloads. The `_einop` method could then be dramatically reduced to directly invoking the transposition-pairwise-reduction-reshape operations with the given arguments.

yhtang:
Fixed by #185.

PR #185, opened 2022-01-29
Precompile einop specifications using live indices. Fixes #182.

### LINKED ISSUE #182: [BUG] Incorrect result using `*` to denote Kronecker product operations
**Describe the bug**
The inner product between a vector and a vector Kronecker product is computed incorrectly.

**To Reproduce**
```
import funfact as ff
u = ff.tensor('u', 5)
v = ff.tensor('v', 2)
w = ff.tensor('w', 3)
tsrex = u[[*i]] * v[[*i]] * w[i]
print(tsrex.shape)
```
leads to
```
()
```

**Expected behavior**
A runtime error should be raised between `w` and `u \otimes v` have incompatible shapes.

**Analysis**
The use of the `|...` syntax in the einspec string to denote Kronecker product indices seems to be the trigger of error here. Specifically, this usage cannot distinguish whether the Kronecker index means '_this index marks an axis to participate in a future Kronecker product_' or '_this index is the result of a Kronecker product_'.

**Proposed solution**
This issue overlaps with #179 because the use of an einspec string seems to introduce unnecessary complications, while also preventing our interpreter to make use of the rich information already carried by the AST. In fact, it should be solved as a by-product as soon as #179 is addressed.

yhtang:
Implements the mechanism as suggested in #182.

### DIFF
--- a/funfact/lang/_tsrex.py
+++ b/funfact/lang/_tsrex.py
@@ -13,7 +13,7 @@ from ._ast import _AST, _ASNode, Primitives as P
 from .interpreter import (
     ASCIIRenderer,
     TypeDeducer,
-    EinsteinSpecGenerator,
+    EinopCompiler,
     IndexnessAnalyzer,
     LatexRenderer,
 )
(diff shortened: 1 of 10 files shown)


yhtang:
Also fixes #179.

PR #188, opened 2022-01-31
Encapsulate each backend as a module instead of a class

yhtang:
Paving the way for #166.


funfact/backend/_proxy.py
```diff
@@ -0,0 +1,153 @@
...
+            try:
+                return getattr(ab.nla, attr)
+            except AttributeError:
+                raise AttributeError(
+                    f'Backend {ab.__name__} does not implement {attr}.'
+                )
+
+    def is_native(self, array):
+        '''Determine if the argument is of type native_t.'''
+        return isinstance(array, self.native_t)
```
campsd, line 117: Should this be `_get_active_backend().native_t` instead of `self.native_t`?

campsd: Same remark applies to the three functions after this one

funfact/backend/_meta.py
```diff
@@ -1,71 +1,49 @@
...
-    def is_tensor(self, array):
-        '''Determine if the argument is one of tensor_t.'''
-        return isinstance(array, self.tensor_t)
-
-    def log_add_exp(self, lhs, rhs):
-        return self.log(self.add(self.exp(lhs), self.exp(rhs)))
-
-    def log_sum_exp(self, data, axis=None):
-        return self.log(self.sum(self.exp(data), axis=axis))
+_required_properties = [
```
campsd, line 4: It seems that this file is no longer used?

funfact/backend/_proxy.py
yhtang: It's actually good since we'll trigger `__getattr__`, which delegates the query to the underlying backend. :grin:

funfact/backend/_meta.py
yhtang: I am going to add a test that uses this list to check if the backends are fully compliant.

campsd:
Looks good to me after the issues are fixed.

PR #192, opened 2022-02-02
Several ML math functions.

yhtang:
ReLU, CeLU, Sigmoid, and replaced our own implementation of log-add-exp with that from the backends.


campsd:
Maybe update the nonnegative condition to the RELU function?

campsd (approved):

PR #199, opened 2022-02-07
First integration tests

campsd:
Some first integration tests that cover a few of the recently fixed bugs.

@yhtang `tox.ini` is changed to the `tests` directory. This runs the tests in that folder, but no longer the unit tests in the `funfact` directory. Can you have a look at this? Maybe a reasonable solution would be to create a `unit` and `integration` subdirectory in `tests` and move all tests there?

### DIFF
--- a/tests/test_tsrex.py
+++ b/tests/test_tsrex.py
@@ -8,22 +8,18 @@ from funfact import active_backend as ab
 
 @pytest.fixture
 def my_arrays():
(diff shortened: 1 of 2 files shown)


yhtang:
I think while I fixed the problem, I just accidentally merged it even though the PR is still a draft. So I'll just close this PR accordingly. The tests are very nicely designed though. Thanks!

PR #211, opened 2022-02-17
Workspace/joss submission

campsd:
Further updates to the paper.

If we switch out the figure with the updated one, we should be in reasonably good shape already.


campsd:
I added the statement of need to the documentation page in this PR, but that is of course not on the correct branch. I'll create another PR targeting the master branch.

PR #213, opened 2022-02-18
Factorization object only return unique factors. Fixes #210.

### LINKED ISSUE #210: [BUG] Duplicate enumeration of factor tensors
**Describe the bug**
If a tensor shows up more than once in a tensor expression, then it will also show up as many times in `Factorization.factors`. This may lead to incorrect behavior during SGD should the same tensor gets updates multiple times during one step.

**To Reproduce**
```
import funfact as ff
a = ff.tensor('a', 4)
fac = ff.Factorization.from_tsrex(a * a)
print(fac.factors)
```
leads to
```
<'data' fields of tensors a, a>
```

**Expected behavior**
A tensor should only appear once in `.factors` regardless of how many times it is used in a tensor expression.

yhtang:


campsd:
Looks good. I like the secret key 😄 

One thing that I noticed, `append` was replaced by `vec_axis` in some but not all places. Any particular reason for having two different names?

yhtang:
> One thing that I noticed, `append` was replaced by `vec_axis` in some but not all places. Any particular reason for having two different names?

Well, it was when I was using it that I realize the `append` argument, without a proper context, may seem unintuitive for users. Hence, I renamed the user-facing arguments to `vec_axis`, but kept all internal usage as `append`. What do you think?

campsd (approved):

campsd (approved):

PR #218, opened 2022-02-28
bugfix #217 Tensor transposition

### LINKED ISSUE #217: [BUG] Tensor transposition
**Describe the bug**
Transposition of individual tensors is failing. The root cause appears to be that this not implemented in the `einop_compiler`.

**To Reproduce**
```py
a = ff.tensor('a', 2, 3, 4, 5)
i, j, k, l = ff.indices('i j k l')
tsrex = a[i, j, k, l] >> [l, i, j, k]
fac()
# >>> TypeError: tran() missing 1 required positional argument: 'einspec'
```

**Expected behavior**
Tensor `a` should be transposed.

campsd:
Closes #217

### DIFF
--- a/funfact/lang/interpreter/_einop_compiler.py
+++ b/funfact/lang/interpreter/_einop_compiler.py
@@ -135,8 +135,7 @@ class EinopCompiler(TranscribeInterpreter):
     def tran(self, src: P.Numeric, indices: P.indices, live_indices, **kwargs):
         return as_namedtuple(
             'transpec',
-            in_idx=src.live_indices,
-            out_idx=indices.live_indices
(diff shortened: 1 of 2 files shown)


yhtang:
Why not we directly store `[transpec.in_idx.index(i) for i in transpec.out_idx]`, so that it don't have to be re-computed again during each evaluation? :grin:

campsd:
> Why not we directly store `[transpec.in_idx.index(i) for i in transpec.out_idx]`, so that it don't have to be re-computed again during each evaluation? 😁

Makes sense 😃  Updated in latest commit.

yhtang (approved):

yhtang:
Thanks!

Issue #225, opened 2022-03-22
[REQ] History of loss function for all vector instances


yhtang:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

PR #226, opened 2022-03-25
Redesign of the gradient descent loop

yhtang:
Uses a driver+plugin pattern to simplify the insertion of user-defined actions to the factorization process. These custom actions/plugins can be used for monitoring/debugging/logging purposes.

Can be used to implement #225.

### DIFF
--- a/funfact/algorithm.py
+++ b/funfact/algorithm.py
@@ -188,7 +188,7 @@ def factorize(
     '''run the gradient descent loop'''
     gradient_descent(
         lambda: loss_and_grad(fac, target), opt, max_steps, exit_condition,
-        plugins=[
+        plugins=plugins + [
             save_best,
             convergence_check
         ]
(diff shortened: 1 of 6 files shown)


funfact/algorithm.py
```diff
@@ -1,18 +1,22 @@
...
+    gradient_descent_plugin,
+    GradientDescentState,
+)
 
 
 def factorize(
     tsrex, target, optimizer='Adam', loss='MSE', lr=0.1, tol=1e-6,
     max_steps=10000, vec_size=1, vec_axis=0, stop_by='first', returns='best',
-    checkpoint_freq=50, dtype=None, penalty_weight=1.0
+    checkpoint_freq=50, dtype=None, penalty_weight=1.0, plugins=[]
```
campsd, line 19: Should we add something about the `plugin` argument to the doc string below?

yhtang: Good point. In fact, the entire plugin system should be documented, probably via a later PR. For this one I'll add a simple one-liner for the time being.

PR #234, opened 2022-04-19
matrix MSE loss agnostic to global phase

campsd:
Loss function that doesn't take into account the global phase of the unitary.

```py
ut, _ = np.linalg.qr(np.random.randn(4, 4) + 1j * np.random.randn(4, 4))
eps = 1e-4
um, _ = np.linalg.qr(ut + eps * (np.random.randn(4, 4) + 1j * np.random.randn(4, 4)))
ut = ab.tensor(ut)
um = ab.tensor(um)
umm = ab.stack([np.exp(1j * np.random.randn()) * um for i in range(5)], dim=2)
ff.loss.mse(um, ut)  # this is small
ff.loss.mse(umm, ut, vectorized_along_last=True, sum_vec=False) # these are big
gp_mse = ff.loss.GlobalPhaseMSE()
gp_mse(umm, ut, vectorized_along_last=True, sum_vec=False) # these are small
```

### DIFF
--- a/funfact/loss.py
+++ b/funfact/loss.py
@@ -134,11 +134,13 @@ class KLDivergence(Loss):
         return ab.multiply(target, ab.log(ab.divide(target, model)))
 
 
-class GlobalPhaseMSE(MatrixLoss):
+class PhaseInvariantMSE(MatrixLoss):
     '''MSE loss for matrices agnostic to global phase.'''
     def _loss(self, model, target):
-        return ab.abs(ab.transpose(model.conj(), axes=(1, 0)) @ target) - \
-               ab.eye(model.shape[0], target.shape[1])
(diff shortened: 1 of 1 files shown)


funfact/loss.py
```diff
@@ -101,6 +134,13 @@ def _loss(self, model, target):
         return ab.multiply(target, ab.log(ab.divide(target, model)))
 
 
+class GlobalPhaseMSE(MatrixLoss):
```
yhtang, line 137: Could we rename it to something like `PhaseInvariantMSE` to make it more intuitive? 😄

yhtang (changes requested):

yhtang (approved):

Issue #221, opened 2022-03-08
Qiskit integration


campsd:
Provide an interface between qiskit and FunFact

- [ ] from qiskit circuit to FunFact tensor expression or intermediate internal circuit representation
- [ ] from FunFact tensor expression or intermediate internal circuit representation to qiskit circuit

PR #235, opened 2022-04-20
funfact quantum module

campsd:
A module for simple quantum circuits and gates.

Example usage:
```py
import funfact.quantum as ffq
circ = ffq.Circuit(3)
circ.append(ffq.TwoQubitUnitary(0))
circ.append(ffq.TwoQubitUnitary(1))
circ.append(ffq.CX(0, 2))
tsrex = circ.to_tensor()
fac()
```

This can help us optimize parametrized circuits, and eventually integrate with qiskit #221

There's two outstanding issues:
1. the `dtype` should be handled in a consistent manner: it should always be a complex data type?
2. the `offset` in the circuits are not yet handled correctly when circuits are added to circuits.


campsd:
Another open issue is how we can map an optimized tensor expression back to a circuit object. This could as simple as iterating over the factors of the optimized model and assigning the result to the tensors in the circuit object. What do you think @yhtang ?

funfact/quantum/_qcircuit.py
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+from numbers import Integral
+from funfact.lang._special import eye
+
+
+class Circuit():
+    '''A class to create quantum circuits and tensor expressions from them.'''
```
yhtang, line 8: This docstring sentence is a bit ambiguous in terms of syntax. :joy:

funfact/quantum/_qcircuit.py
```diff
@@ -0,0 +1,49 @@
...
+            raise RuntimeError(
+                f'Offset must be integer >= 0, got {offset}.'
+            )
+
+        self._nbqubits = nbqubits
+        self._offset = offset
+        self._qgates = []
+
+    @property
+    def nbqubits(self):
```
yhtang, line 24: What does this property do? Could we add a docstring for it?

funfact/quantum/_qgates.py
```diff
@@ -0,0 +1,225 @@
...
+            if not (isinstance(q, Integral) and q >= 0):
+                raise RuntimeError(
+                    f"Qubit indices must be non-negative integers, got {q} for"
+                    f" entry {i}."
+                )
+        qubits.sort()
+        self._qubits = qubits
+
+    @abstractmethod
+    def to_tensor(self):
```
yhtang, line 32: Maybe `to_tsrex` to make it clear this creates an abstract expression?

funfact/quantum/_qgates.py
```diff
@@ -0,0 +1,225 @@
...
+    '''A common base class for all gate objects.'''
+    def __init__(self, qubits: List[Integral]):
+        '''A gate stores the minimum and maximum qubits it acts on.'''
+        for i, q in enumerate(qubits):
+            if not (isinstance(q, Integral) and q >= 0):
+                raise RuntimeError(
+                    f"Qubit indices must be non-negative integers, got {q} for"
+                    f" entry {i}."
+                )
+        qubits.sort()
```
yhtang, line 28: `.sort()` modifies the input argument in-place and may cause unexpected behavior in a complex program. A safer option is to use `sorted` and store the result as a `tuple`, which is immutable.

funfact/quantum/_qgates.py
```diff
@@ -0,0 +1,225 @@
...
+        '''A gate can be converted to a tensor, either parametrized or not.'''
+        pass
+
+    @abstractmethod
+    def label(self):
+        '''A gate has a label.'''
+        pass
+
+    @property
+    def qubits(self):
```
yhtang, line 42: Is there a way that we can hint this is a list of 0-based indices, i.e. the ids of the qubits?

funfact/quantum/_qgates.py
```diff
@@ -0,0 +1,225 @@
...
+        return self._qubits
+
+
+class OneQubitGate(Gate):
+    '''Abstract base class for gates acting on exactly one qubit.'''
+    def __init__(self, qubit: Integral):
+        if not (isinstance(qubit, Integral)):
```
yhtang, line 49: Since we have this type of integral check in multiple places in the PR, we can factor them out as a utility method. Alternatively, since the requirement of qubit ID being integer is relatively trivial and intuitive, we can simply mention it in the docstring and then forgo the asserts.

funfact/quantum/_qgates.py
```diff
@@ -0,0 +1,225 @@
...
+        super().__init__(qubit)
+
+    def to_tensor(self):
+        return tensor(
+            self.label, 2, 2, initializer=self._initializer, prefer=Unitary()
+        )
+
+    @property
+    def label(self):
+        return 'Q_1'
```
yhtang, line 69: Perhaps add an optional argument in the constructor to let users customize the label?

funfact/quantum/_special.py
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+from funfact.lang._ast import Primitives as P
+from funfact.lang._terminal import AbstractTensor
+from funfact.lang._tsrex import TsrEx
+from funfact import active_backend as ab
+import numpy as np
+
+
+def proj0(dtype=None):
```
yhtang, line 10: docstrings?

funfact/quantum/_special.py
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env python
```
yhtang, line 1: How about merging this file with `lang._special`?

yhtang (changes requested):

yhtang:
Nice work overall!

yhtang:
> There's two outstanding issues:
> 
> 1. the `dtype` should be handled in a consistent manner: it should always be a complex data type?

Agree. For this particular application area, I guess we can start with the two most common options `complex64` and `complex128`. Not sure if 16-bit floats is useful but it could the subject of a future project :grin:

> 2. the `offset` in the circuits are not yet handled correctly when circuits are added to circuits.

Could you please clarify a little bit? Does `offset` refer to the position of the qubits?

yhtang:
> Another open issue is how we can map an optimized tensor expression back to a circuit object. This could as simple as iterating over the factors of the optimized model and assigning the result to the tensors in the circuit object. What do you think @yhtang ?

This can be break down into two sub-problems that we can solve independently:
1. Is there a way for us to uniquely label each circuit & its corresponding abstract tensor?
2. Is there a way to enumerate/access the gates in a circuit so that we can update them with concrete tensors?

funfact/quantum/_qgates.py
campsd: Do you mean hint in the doc string?

funfact/quantum/_qgates.py
campsd: I've added the documentation but kept the check in the `util` module. Let me know if you'd rather remove it.

funfact/quantum/_qcircuit.py
campsd: It is the number of qubits (or size) of the circuit. Included a doc string.

funfact/quantum/_special.py
campsd: Done

campsd:
> > There's two outstanding issues:
> > 
> > 1. the `dtype` should be handled in a consistent manner: it should always be a complex data type?
> 
> Agree. For this particular application area, I guess we can start with the two most common options `complex64` and `complex128`. Not sure if 16-bit floats is useful but it could the subject of a future project 😁

I was unsure how we enforce this, but I guess using something like `fac = ff.Factorization.from_tsrex(circ.to_tsrex(), dtype=ab.complex128)` would work?
> 
> > 2. the `offset` in the circuits are not yet handled correctly when circuits are added to circuits.
> 
> Could you please clarify a little bit? Does `offset` refer to the position of the qubits?

This optional argument specifies if the qubits of the circuit don't range from `0` to `nbqubits-1` but from `offset` to `nbqubits-1+offset`. It is ignored if a circuit only consists of gates, but if you add a circuit to another circuit, you can control the placement of the subcircuit with this. For example:

```py
circ2 = ffq.Circuit(2, offset=0)
circ2.append(ffq.OneQubitUnitary(0))
circ2.append(ffq.CX(1, 0))
circ1 = ffq.Circuit(1, offset=1)
circ1.append(ffq.PauliY(0))
circ2.append(circ1) # circ1 now acts on qubit 1 of circ2
circ2.to_tsrex()
```

campsd:
> > Another open issue is how we can map an optimized tensor expression back to a circuit object. This could as simple as iterating over the factors of the optimized model and assigning the result to the tensors in the circuit object. What do you think @yhtang ?
> 
> This can be broken down into two sub-problems that we can solve independently:
> 
> 1. Is there a way for us to uniquely label each circuit & its corresponding abstract tensor?
> 2. Is there a way to enumerate/access the gates in a circuit so that we can update them with concrete tensors?

This makes sense. Shall we handle this in a follow up PR?

funfact/quantum/_qgates.py
campsd: I've added a hint in the doc string now

campsd:
Also for future reference: the `to_tsrex` function in the `Circuit` class is very naive. It places one gate and blows it up to dimension `2**n`, this can be significantly improved to reduce the cost. I suggest that we move forward with the naive implementation for now and optimize it later.

yhtang:
> This makes sense. Shall we handle this in a follow up PR?

Sure. I'm pretty sure this is going to take more than one PRs to accomplish :grin:

yhtang:
> I was unsure how we enforce this, but I guess using something like `fac = ff.Factorization.from_tsrex(circ.to_tsrex(), dtype=ab.complex128)` would work?

Actually this sounds like a better option. So basically we factor out the representation of a quantum circuit from its numerical realization.

yhtang:
> This optional argument specifies if the qubits of the circuit don't range from `0` to `nbqubits-1` but from `offset` to `nbqubits-1+offset`. It is ignored if a circuit only consists of gates, but if you add a circuit to another circuit, you can control the placement of the subcircuit with this. For example:
> 
> ```python
> circ2 = ffq.Circuit(2, offset=0)
> circ2.append(ffq.OneQubitUnitary(0))
> circ2.append(ffq.CX(1, 0))
> circ1 = ffq.Circuit(1, offset=1)
> circ1.append(ffq.PauliY(0))
> circ2.append(circ1) # circ1 now acts on qubit 1 of circ2
> circ2.to_tsrex()
> ```

I see. Would it be better to specify the placement of a circuit at the time when it gets inserted into another one? In this way, we can define a circuit once and then reuse it multiple times at different places in a bigger circuit.

This feels a bit like the index renaming mechanism that we implemented to the tensor expressions. For example, we can define a circuit using a set of 'nomial' qubits and then map them onto actual ones when placing them into a bigger one.

yhtang:
> This feels a bit like the index renaming mechanism that we implemented to the tensor expressions. For example, we can define a circuit using a set of 'nomial' qubits and then map them onto actual ones when placing them into a bigger one.

I wonder how qiskit or a hardware description language handles this scenario.😎

campsd:
> > This feels a bit like the index renaming mechanism that we implemented to the tensor expressions. For example, we can define a circuit using a set of 'nomial' qubits and then map them onto actual ones when placing them into a bigger one.
> 
> I wonder how qiskit or a hardware description language handles this scenario.😎

For how qiskit handles this scenario, see: https://qiskit.org/documentation/tutorials/circuits_advanced/01_advanced_circuits.html#Composition

you can attach a subcircuit to certain qubits of the main circuit in the `append` function.

funfact/quantum/_qgates.py
```diff
@@ -0,0 +1,264 @@
...
+from funfact.lang._ast import Primitives as P
+from funfact.lang._terminal import ParametrizedAbstractTensor
+from funfact.lang._special import eye, proj0, proj1
+from funfact.conditions import Unitary
+from funfact.util.integral import check_bounded_integral
+
+
+class Gate(ABC):
+    '''A common abstract base class for all gate objects.'''
+    def __init__(self, qubits: List[Integral]):
```
yhtang, line 19: Delayed binding of a gate to qubits via `at`/`@`.

yhtang (changes requested):

campsd:
Further updates based on our discussion before. 

Current usage examples:

```py
circ = ffq.Circuit(3)
G1 = ffq.TwoQubitUnitary()
circ.append(G1 @ (0, 1))
circ.append(G1, at=(1, 2))
```

```py
circ = ffq.Circuit(2)
CNOT = ffq.CX()
circ.append(CNOT @ (1, 0)) # first = control, second = target
```

yhtang (approved):

